PR #5, opened 2019-04-08
🔥 remove bower and gulp, update jest and eslint

eliascodes:
These are all unnecessary and play no role in the build process.


astroash:
Manual testing
- complete local build

PR #13, opened 2019-04-11
✨ Utilities package

eliascodes:
Extracting common utilities out of the API and into a lib directory. Generic operations on native data structures should go here.


api/src/services/pdf/templates/__tests__/visitor_qr_code.test.unit.ts
```diff
@@ -26,13 +25,14 @@ describe('Visitor QR Code Template', () => {
...
 
-    test('Create template w/ custom logo as data URL', async () => {
+    test('Create template w/ custom logo as image URL', async () => {
       const imageContents = '20t48hwfbekjaofhw08rq3hiofnw';
       const url = 'https://example.com/images/fake.png';
       const qrCodeDataUrl = 'data:image/png;base64,fooqfoi2oitn';
 
-      const mock = new MockAdapter(axios);
-      mock.onGet(url).reply(200, imageContents);
+      nock('https://example.com')
```
astroash: Do you know why this works with nock and not axios-mock-adapter?

PR #15, opened 2019-04-12
🐛 fix formik test woes

astroash:
- wrap expects in additional async/wait functions
https://github.com/TwinePlatform/volunteer-dashboard/issues/8


eliascodes:
does this work without `--runInBand`?

PR #16, opened 2019-04-12
♻️ Api/email refactor

### LINKED ISSUE #73: Make domain part of the email template model
Annoying not being able to test e-mail related features on staging without changing the templates on postmark. The domains should be part of the config.

eliascodes:
Fixes #73 

### Changes
* Add low-level `Dispatcher` interface which handles calls to third party API (Postmark)
* Add high-level `Service` interface whose API reflects domain-concepts and which takes care of making the appropriate calls to the `Dispatcher`
* Add tests :)
* Add mock simple for `Postmark` module
* Replace calls to old `EmailService` with calls to new interface
* Update existing tests/mocks


api/src/services/email/index.ts
```diff
@@ -1,65 +1,132 @@
...
+};
 
-const emailInitialiser: EmailInitialiser = {
-  init: (config: EmailServiceConfig) => {
-    const emailClient = new Postmark.ServerClient(config.apiKey);
 
-    return {
-      send: async (emailOptions: Email) => {
-        const postmarkOptions = emailKeyMap(emailOptions);
+const Service: EmailService = {
```
astroash: Why's this called `Service` rather than email service? If we add alternative services to our api this will get confusing.

eliascodes: I called it that to disambiguate between the type and the actual object. The name is internal to the module only. When it's imported you can call it whatever you like.

PR #26, opened 2019-04-18
✨ setup pages for new dashboard

### LINKED ISSUE #25: ✨ Setup page structure for dashboard
Designs are not currently signed off, but we can make container pages for the dashboard that call the relevant data.

astroash:
- add pages to routing
- pages call relevant data
- basic title tests
- navbar shows active page

fixes #25


astroash:
> Are we just not doing those right now in the interests of speed?

Yup. We need some clarification on the navbar eg what it looks like when username is clicked. For now I have just added a logout button for dev purposes.

PR #27, opened 2019-04-18
⬇️ fix typescript compile time

astroash:
- known bug with styled-component types that slows down compile time


astroash:
🚨 self merging

PR #28, opened 2019-04-19
💅 implement basic styles

### LINKED ISSUE #24: 💅 Implement basic design for dashboard
Designs are not signed off yet, but we can start with implementing a basic version of them for now ✨

astroash:
- add style variables as enums
- remove old styling
- update existing pages to look mostly like designs
- add the basis of a navbar

fixes #24


dashboard-app/src/styles/index.css
```diff
@@ -1,11 +1,59 @@
-@import url('https://fonts.googleapis.com/css?family=Nunito:300,400,700');
+@import url('https://fonts.googleapis.com/css?family=Roboto:100,300,400');
 
 * {
-  box-sizing: border-box;
```
eliascodes: How come we're removing this?

dashboard-app/src/components/Navbar/Navigation.tsx
```diff
@@ -0,0 +1,44 @@
...
+  background-color: transparent;
+  font-size: inherit;
+  display: inherit;
+  font-weight: inherit;
+`;
+
+const Navigation: React.FunctionComponent<Props> = (props) => (
+
+  <Row between={'xs'}>
+      <Link to={'./activites'}>Activity</Link>
```
eliascodes: These look like filepaths instead of urls, is that deliberate?

dashboard-app/src/styles/index.css
astroash: Assumed this was added in the css reset, but it's not. Good spot

dashboard-app/src/components/Navbar/Navigation.tsx
astroash: Nope

PR #30, opened 2019-04-23
✨ add Error Page

### LINKED ISSUE #29: 💅 style error page like designs
- add designs
- change copy based on error type

astroash:
- change copy based on params
- add default styling to buttons

fixes #29

NOTE: briefly looked into testing but it is complicated as the component relies on url params. As this component is not vital business logic it doesn't seem worth spending time on.

### DIFF
--- a/dashboard-app/src/Error.tsx
+++ b/dashboard-app/src/Error.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import { RouteComponentProps, withRouter } from 'react-router-dom';
-import { Grid, Row, Col } from 'react-flexbox-grid';
+import { Grid, Row } from 'react-flexbox-grid';
 import { SpacingEnum } from './styles/style_guide';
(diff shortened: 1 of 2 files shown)


dashboard-app/src/Error.tsx
```diff
@@ -67,7 +67,7 @@ const H1 = styled(_H1)`
   margin-bottom: ${SpacingEnum.medium};
 `;
 
-const H3Padded = styled(H3)`
+const H3 = styled(_H3)`
```
astroash: This changes the spacing between elements. I've setup the `SpacingEnum` to follow the design principle of spaces being multiples of each other.

eliascodes: I just changed the name of the variable, the spacing is the same as what you set it as.

astroash: There are 2 H3 elements. Previously only one had padding attached to it, now both do.

eliascodes: Oops, my bad. Will fix.

PR #31, opened 2019-04-24
✨ Utilities package (second attempt)

### LINKED ISSUE #22: 🚀 heroku setup does not support lerna bootstrap
Current heroku setup does not support lerna. In `api` sibling `twine-util` package is unable to install as heroku uses `npm i` on install which returns a `404` for this package.

Steps taken:
- looked into whether we can easily overwrite the install script. Seems like the only way to do this is to edit the node buildpack which I think is a terrible idea
- preinstall lerna bootstrap gets caught in a loop
- postinstall is useless

A number of people seem to be using docker with heroku/learner which may be a solution. Another gave up and deployed with now 😅

~~This is being temporarily fixed by installing the package from a separate repo on gh.~~

Reverting utils refactor to be readded when this issue is solved.

eliascodes:
Extracting common utilities out of the API and into a lib directory. Generic operations on native data structures should go here.

Also use local path to specify this dependency instead of using `lerna bootstrap`.

Fixes #22


astroash:
@eliascodes is this awaiting review?

eliascodes:
Yup, forgot to move it to the right column.

astroash:
@eliascodes have you run a test build with this on heroku? Ideally I would like to avoid merging potentially breaking changes to master again. 

Also it's worth updating the docs to describe the install method used in this PR.

eliascodes:
I deployed to `twine-api-test`. Will update docs.

PR #35, opened 2019-04-30
[WIP] Design system

eliascodes:
Just an FYI PR.


astroash:
@eliascodes can this be closed?

eliascodes:
Yes and no. If you've taken it for a spin, had a flick through the storybook and begun to form an opinion, then it's probably fulfilled its purpose for now.

astroash:
@eliascodes please review and merge

astroash:
@eliascodes thoughts on ejecting this and using babel to compile the react/typescript? Storybook is building it's own version to display but we need to be able to import the js/react components to other repos.

eliascodes:
Yeah I'd prefer not to, but I can't find an obvious solution with `react-scripts`, so sure.

astroash:
@eliascodes I'll add the release ribbon to the apps in a separate PR.

api/config/types.ts
```diff
@@ -7,7 +7,7 @@ import { AppEnum } from '../src/types/internal';
 
 export enum Environment {
   DEVELOPMENT = 'development',
-  TESTING = 'testing',
+  TEST = 'test',
```
eliascodes: Why are we doing this?

lib/twine-ui/src/components/Buttons.tsx
```diff
@@ -0,0 +1,57 @@
+import styled from 'styled-components';
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent components in the dashboard?

lib/twine-ui/src/components/Card.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent components in the dashboard?

lib/twine-ui/src/components/Input.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent components in the dashboard?

lib/twine-ui/src/components/Select.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent components in the dashboard?

lib/twine-ui/src/components/Typography.tsx
```diff
@@ -0,0 +1,52 @@
+import styled from 'styled-components';
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent components in the dashboard?

lib/twine-ui/src/styles/colours.ts
```diff
@@ -0,0 +1,51 @@
+export enum ColoursEnum {
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent files in the dashboard?

lib/twine-ui/src/styles/fonts.ts
```diff
@@ -0,0 +1,34 @@
+export enum FontFamilyEnum {
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent components in the dashboard?

lib/twine-ui/src/styles/spacing.ts
```diff
@@ -0,0 +1,3 @@
+export enum PaddingEnum {
```
eliascodes: This was added a while ago, is this still consistent with (1) any updates to the design system and (2) the equivalent components in the dashboard?

api/config/types.ts
astroash: Because this storybook expects this variable to be set in testing environment

lib/twine-ui/src/components/Buttons.tsx
astroash: Im happy to pull out all the old components and readd them as we slowly move over to this library

astroash:
> Also where did we land on ejecting

More apt solution:
- remote CRA
- compile with typescript

lib/twine-ui/src/components/Buttons.tsx
eliascodes: Yeah lets do that.

eliascodes:
> remove CRA

OK but `react-scripts` is still installed here?

astroash:
Benching as this is more work than initially thought. There's a number of compile issues to do with using styled components. Need to look into webpack & babel options.

Issue #39, opened 2019-04-30
💅 create tables for dashboard


astroash:
- [x] create table component
- [x] add sorting and filtering functionality
- [x] style it

PR #41, opened 2019-04-30
💅 Add date picker

eliascodes:
Related to #39 

### Changes
* Install `react-datepicker`
* Overwrite `react-datepicker` styles with custom ones to match initial design
* Separate definition for `Label` component
* Allow custom select box to accept child elements
* Group font related enums into one object

No unit tests here because the main dependency of `react-datepicker` doesn't work nicely with `jsdom` (see https://github.com/FezVrasta/popper.js/issues/478)


dashboard-app/src/components/DatePicker/styles.css
```diff
@@ -0,0 +1,22 @@
+.react-datepicker__header {
+  background-color: white;
```
astroash: Can this be a `ColourEnum`?

dashboard-app/src/components/DatePicker/styles.css
```diff
@@ -0,0 +1,22 @@
...
+  width: 90%;
+  margin: 0 auto;
+}
+
+.react-datepicker__month {
+  margin: 0.2em;
+}
+
+.react-datepicker__month--selected {
+  background-color: #707070;
```
astroash: `ColourEnum`

dashboard-app/src/components/DatePicker/styles.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.react-datepicker__month-text.react-datepicker__month--selected:hover {
+  background-color: rgba(112, 112, 112, 0.8);
```
astroash: `ColourEnum`

dashboard-app/src/components/Select.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+`;
+
+export const Option = styled.option`
+  font: ${Fonts.family.default};
+  font-size: ${Fonts.size.normal};
+`;
+
+const ErrorText = styled.span`
+  color: red;
```
astroash: `ColourEnum`

astroash:
Can you also check console errors? Not sure whether mine are caused my faux props:

Screenshot 2019-05-01 at 10 45 50

Screenshot 2019-05-01 at 10 45 34

dashboard-app/src/components/DatePicker/styles.css
eliascodes: Its a CSS file.

dashboard-app/src/components/DatePicker/styles.css
eliascodes: Its a CSS file

eliascodes:
> Can you also check console errors? Not sure whether mine are caused my faux props

I don't get those errors, I just get warnings about iterated elements not having keys (which I've just fixed in the most recent commit). I also get a warning about the `isActive` prop which is something from the navbar, so I wont address that here.

PR #42, opened 2019-04-30
✨ Add utility bar

eliascodes:
Related to #39

### Changes
* Styling fixes to `DataTable` component
* Add `Toggle` and `UnitToggle` button components
* Add `UtilityBar` component

### DIFF
--- a/dashboard-app/src/components/DataTable/DataTableHeaderRow.tsx
+++ b/dashboard-app/src/components/DataTable/DataTableHeaderRow.tsx
@@ -12,7 +12,10 @@ const HeaderRow = styled.tr`
 `;
 
 const HeaderCell = styled.th`
-  padding: ${SpacingEnum.small} 27px;
+  padding-top: ${SpacingEnum.small};
+  padding-bottom: ${SpacingEnum.small};
+  padding-left: ${SpacingEnum.small};
+  padding-right: calc(27px + calc(27px - ${SpacingEnum.small}));
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
--- a/dashboard-app/src/components/DataTable/DataTableHeaderRow.tsx
+++ b/dashboard-app/src/components/DataTable/DataTableHeaderRow.tsx
@@ -15,7 +15,7 @@ const HeaderCell = styled.th`
   padding-top: ${SpacingEnum.small};
   padding-bottom: ${SpacingEnum.small};
   padding-left: ${SpacingEnum.small};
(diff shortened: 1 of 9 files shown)


eliascodes:
This is the last piece needed before we can start arranging things on the pages themselves and hooking up the data.

dashboard-app/src/components/DataTable/DataTableHeaderRow.tsx
```diff
@@ -12,7 +12,10 @@ const HeaderRow = styled.tr`
 `;
 
 const HeaderCell = styled.th`
-  padding: ${SpacingEnum.small} 27px;
+  padding-top: ${SpacingEnum.small};
+  padding-bottom: ${SpacingEnum.small};
+  padding-left: ${SpacingEnum.small};
+  padding-right: calc(27px + calc(27px - ${SpacingEnum.small}));
```
astroash: what's going on here?

dashboard-app/src/components/UnitToggle.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Toggle from './Toggle';
+
+
+type DurationUnit = 'Hours' | 'Days';
```
astroash: I've defined this as a `UnitEnum` on my current branch. Wanna do the same here, then it can be refactored in mine. I can't think of any benefits of it being a union type, and an enum will allow for it to be easily passed into copy.

dashboard-app/src/components/UtilityBar.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useState, useCallback } from 'react';
+import moment from 'moment';
+import { Row, Col } from 'react-flexbox-grid';
+import DatePicker from './DatePicker';
+import UnitToggle from './UnitToggle';
+
+
+type DurationUnit = 'Hours' | 'Days';
```
astroash: `UnitEnum`

dashboard-app/src/components/DataTable/DataTableHeaderRow.tsx
eliascodes: We want the padding between the cells to be 54px, but we also want the left padding to be 1em to match the data cells, so this just works out `54px - 1em`, which is probably what this should be changed to

astroash:
Screenshot 2019-05-03 at 11 14 44

To changes:
- To & From boxes are different sizes
- border to the unit toggle does not match dropdown border

astroash:
Note: To box is still larger. I think the sizing should be on the dropdown box itself rather than taking up the remaining space in the parent element. This can be resolved later

PR #46, opened 2019-05-02
♻️ Refactor the navigation elements

eliascodes:
### Changes
* Add central `Pages` abstraction which deals with the relationship b/w URLs, components, and display titles
* Add utility functions to deal with `Page` data type
* Update `Navbar` and `App` components to use new abstraction
* Move nav stuff into `navigation` directory
* Add tests :)


dashboard-app/src/navigation/pages.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    if (keys.length !== 1) {
+      throw new Error(`One page expected, ${keys.length} returned: ${keys}. ${pathname}`);
+    }
+
+    return pages[keys[0]];
+  },
+
+  toRoute (page: Page) {
+    return <Route exact path={page.url} component={page.component} key={page.url} />;
```
astroash: This is assumes that we will always be rendering pages identically, removing our ability to use alternative methods available with `react-router-dom`. We should be careful with putting abstraction layers on top of frameworks as it ties us into their current api and doesn't leave room for flexibility. 

I think abstraction on top of our own domain logic, but IMO the abstraction in this PR goes a little too far.

eliascodes: yeah fair, this is mostly a utility method, but suppose it can just live in `App.tsx` instead

astroash:
Screenshot 2019-05-03 at 11 14 44

To changes:
- To & From boxes are different sizes
- border to the unit toggle does not match dropdown border

dashboard-app/src/navigation/Navbar/index.tsx
```diff
@@ -28,30 +34,25 @@ const Title = styled.p`
...
   return(
   <PaddedRow middle="xs" between="xs">
     <Col xs={6} lg={4}>
       <Title>TWINE</Title>
     </Col>
     <Col xs={6} lg={4}>
-    {isLoggedIn &&
-      <Navigation active={active}/>
-    }
+      { isLoggedIn && <NavLinks links={links} withLogout={isLoggedIn}/> }
```
astroash: Is there a page where we want the navlinks showing but not the logout link?

eliascodes: The designs have navbar links on the login page

astroash: The designs were not meant to be a UX design in this sense. And I think we should be careful coding in specifics that are shown. 

It doesn’t matter too much for this as it’s a small change, but worth thinking about as we develop the rest of the dashboard.

PR #52, opened 2019-05-08
📊 more dashboard table work

astroash:
- months column headers display year
- tests for table manipulation helper functions
- error handling for data manipulation
- simple display for messages (requests and data manipulation)


dashboard-app/src/hooks/useRequest.ts
```diff
@@ -1,16 +1,34 @@
...
   params?: any
   updateOn?: any[]
   callback?: (d: any) => void
+  setErrors: (d: any) => void
+  push: any
 };
 
-export default ({ apiCall, params = {}, updateOn = [], callback }: RequestParams) => {
-  const [error, setError] = useState<string>();
+const onError = (error: any, setErrors: any, push: any) => {
```
eliascodes: The types aren't that involved here, we may as well stick them in:
`(error: AxiosError, setErrors: (d: Dictionary<string>) => void, push: (p: string) => void)`

eliascodes:
This can go in first probs.

dashboard-app/src/hooks/useRequest.ts
astroash: `AxiosError` wasn't playing nicely with the Response utils, as it doesn't have all the same properties as `AxiosResponse`. I've left this one for now.

Issue #48, opened 2019-05-03
💅 Util bar feedback 


astroash:
- [ ] to/from dropdown boxes are different sizes
- [x] start date should not be restricted. Only allow a maximum of 12 months (eg if I pick June 2016 I can only pick up to May 2017)

PR #53, opened 2019-05-08
Integrate design system

eliascodes:
Related to #48

Most major visual issues and bugs should have been taken care of in this PR. Some minor stuff might be left, but IMO should be dealt with separately in the interests of not making merge conflicts horrible for one or both of us.

### Changes
* Set root font size based on window width
* Use `rem` units everywhere (so that everything is styled in proportion to the root font size)
* Add a simple initial Footer
* I haven't pulled anything into `SpacingEnum` mainly because I'm not sure there's a good way of categorising all the spacing yet (I haven't looked super closely at the designs to see if there's a reasonable pattern, but I suspect not)
* Beyond that, all spacing is modeled directly off the designs
* The `sortBy` state is now kept in the parent instead of in the `DataTable` to allow dynamic changing of the column header text
* Remove `overflow: hidden` on `body` element 🙈
* Fix restrictions on date-picker: "from" is free-select month, "to" is max 12 months after, 1-month minimum (i.e. I can select Jan 2019 - Jan 2019 to view the whole month).

### Open issues to be dealt with separately
* Navbar links are lower-case except for Logout
* Navbar doesn't contain a "Home" button
* Navbar is empty when not logged in
* On reflection, table columns should probably just be fixed width


astroash:
> Navbar is empty when not logged in

This is currently meant to be the case

astroash:
Styling feedback:
Screenshot 2019-05-08 at 17 40 22
Navbar 
- [ ] "TWINE" should not be bold
- [ ] link text is too small

Page
- [ ]  footer should stick to bottom

astroash:
Styling Feedback:
Screenshot 2019-05-08 at 17 46 24

Tables

- [ ] tables take up too small a portion of the screen, can we wider
- [ ] too much spacing before page header

astroash:
Screenshot 2019-05-08 at 17 49 29

- [ ] background colour of unit toggle no longer matches border

eliascodes:
>  link text is too small

How big should it be?

astroash:
> > link text is too small

> How big should it be?

The navbar & error pages should already be styled to match the designs. It's the tables that are too big.

eliascodes:
> background colour of unit toggle no longer matches border

It's that dark in the designs

eliascodes:
> The navbar & error pages should already be styled to match the designs. It's the tables that are too big.

The designs have a navbar with `height: 4.8125rem`, links with `font-size: 1rem`, and from the looks of it, `line-height: 1.3125rem` (because the spacing top and bottom of the link text is `1.75rem`). These are the values I've used in the styling.

These `rem` values are calculated from the original design based on a root font-size of `16px` and a screen width of `1920px`. The root font-size is scaled down proportional to the screen size, so on my `1280px` macbook, it's `10.667px`.

So my question is: navbar links are currently `font-size: 1rem`. What font-size should they be?

astroash:
> What font-size should they be?
They're currently `16px` on `master` and display like the designs on my `1280px` screen.

I am not sure if base styling can be done in such mathematical way 🤷🏽‍♀️

eliascodes:
`16px` is `~1.5rem`, which looks like this on my screen:
Screenshot 2019-05-08 at 17 24 24

astroash:
> It's that dark in the designs

The border and fill colour need to be the same. Otherwise it looks janky

eliascodes:
Think I'm done now @astroash

eliascodes:
Worth a quick run-through again to check the rebase didn't mess up the styling (doesn't look like it to me though)

PR #57, opened 2019-05-10
📈 Codeclimate

### LINKED ISSUE #1: Replace codecov with codeclimate
The codecov website is really bad/buggy and is not that useful for code quality analysis.

Codeclimate looks much better.

eliascodes:
Fixes #1 

Add initial setup for codecov

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ services:
   - postgresql
 
 install:
-  - npm install -g codecov
+  - bash ./.travis/cc-download.sh
   - npm install -g lerna
   - lerna bootstrap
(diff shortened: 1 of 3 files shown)


astroash:
@eliascodes do you know what's up with the failing check on codeclimate here?

eliascodes:
Yeah there's loads of code quality issues! 😄 

No, most of them are in vendor and library files (i.e. the dependencies and shit from the volunteer app that we need to that we need to keep in source control). I've added it all to the exclude list so hopefully next time it should only report actual issues to us.

We can also ignore the code-climate issues until we've fine tuned what they actually look for (to avoid flagging stuff that is not really an issue).

PR #60, opened 2019-05-13
💄 fix horizontal scrolling issue

eliascodes:
Lock horizontal scrolling on app.


astroash:
@eliascodes anymore details on what this is fixing?

eliascodes:
When we did the demo on Monday I noticed the app was scrolling horizontally.

PR #72, opened 2019-05-15
✨ add download to csv functionality

### LINKED ISSUE #74: ⬇️ download button
- download data as csv

### LINKED ISSUE #76: Rounding leading to errors in summation
The way that the aggregation of log duration is done at the moment leads to an accumulated error, regardless of unit.

https://github.com/TwinePlatform/twine-monolith/blob/bb1ce8441a5d95cd84cf3d9dcea2eeed4d0e8fd9/dashboard-app/src/util/tableManipulation.ts#L7-L24

Rounding after summing _each log_ onto the previous sum introduces an error that accumulates over the entire iteration. Rounding is an issue _presentational_ logic, and should not affect the accuracy of the summation.

This is one of the problems that arise from having tightly coupled logic (see #56)

astroash:
**Changes**
- refactor data aggregation to separate aggregation & presentation for csv/tables
- add custom hook for aggregating & storing data in state
- add download to csv functionality

fixes #74 
fixes #76


eliascodes:
In this refactor, can you explain how the data transformation code is structured conceptually?

astroash:
Screenshot 2019-05-20 at 12 53 58

(1) resolved ~Download button height isn't the same as the toggle bar.~

(2) hacky fix for now ~the "Total ${unit}" column ends up at the end of the generated CSV which causes a bit of dissonance when comparing to the original table. The columns should ideally be in the same order.~

(3) let's make an issue about this one

- The rows are not in the same order as in the table. Ideally should be in the same order.

PR #90, opened 2019-05-20
✨ abbreviate months

### LINKED ISSUE #81: 📆 Change month format
Abbreviate months to  Jan, Feb, Mar etc. in column headers

astroash:
**Changes:**
- separate formatting of months.
  - agg data is verbose format eg June 2018
  - table data is abbreviated format eg Jun 18

fixes #81

### DIFF
--- a/dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
+++ b/dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
@@ -1,11 +1,27 @@
-import { evolve, map, pipe, merge } from 'ramda';
+import { evolve, map, pipe, merge, toPairs, fromPairs } from 'ramda';
+import moment from 'moment';
 import { DataTableProps } from '../../components/DataTable/types';
 import { AggregatedData } from './logsToAggregatedData';
+import Months from '../months';
 
 interface Params {
   title: string;
   data: AggregatedData;
 }
+
+export const isDateString = (x: any): boolean => {
+  if (typeof x !== 'string') return false;
+  const words = x.split(' ');
+  if (words.length !== 2) return false;
+  if (isNaN(Number(words[1]))) return false;
+  return Months.list.includes(words[0]);
+};
+
+export const abbreviateDateString = (x: string): string => {
+  return moment(new Date(x)).format(Months.format.table);
+};
+
+const abbreviateIfDateString = (x: any): string => isDateString(x) ? abbreviateDateString(x) : x;
 
 const addContentObjects = evolve({
   rows: map(map((y: any) => ({ content: y }))),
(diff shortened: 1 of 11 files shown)


dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
```diff
@@ -1,12 +1,28 @@
...
 import { DataTableProps } from '../../components/DataTable/types';
 import { AggregatedData } from './logsToAggregatedData';
+import Months from '../months';
 
 interface Params {
   title: string;
   data: AggregatedData;
 }
 
+export const isDateString = (x: any): boolean => {
```
eliascodes: Why not:
```ts
export const isDateString = (x: any): boolean => moment(x).isValid()
```

eliascodes:
So RE: what we spoken about on slack:

If I understand your question correctly, the problem is actually that the representation of "month" in the data structure chosen to represent the "aggregate data" isn't the same as the representation of "month" in the data structure required by the presentation layer.

This isn't inherently a problem. In fact, it's probably more desireable to decouple the aggregate data structure from the exact representations required by the presentation layer.

It means that we need an additional conversion when going from "aggregate data" -> "table data" that transforms the "aggregate" representation of month to the "table" representation. Basically we need another `moment(month).format(newFormat)` in `aggregatedToTableData`.

eliascodes:
On a separate note, I would suggest that including a `Total ${unit}` column in the aggregated data means its too closely coupled to the table (rather than the graph, which won't have that data-point).

eliascodes:
Another suggestion is to compose `useRequest` with the hook that calculates aggregated data:

```ts
import React, { useEffect, useState } from 'react';
import { useRequest } from '../hooks';

const useAggregatedData = (updateOn) => {
  const [rawData, setRawData] = useState(); // optional I guess
  const [aggData, setAggData] = useState();
  const [error, setError] = useState();

  useEffect(() => {
    const { data, error, loading } = useRequest(CommunityBusinesses.getLogs);
    // ...
  }, updateOn);

  useEffect(() => {
    setAggData(logsToAggregatedData(aggData));
  }, [rawData]);

  return {
    aggregatedData: aggData,
    error,
  };
};

// ...

const ByActivity = () => {
  // ...
  const { aggregatedData, error } = useAggregatedData([fromDate, toDate]);
  // ...
}
```

eliascodes:
Final comment on naming of hooks, it seems as though the convention so far is to name the `use<NOUN>`, rather than `use<VERB>`, so I think `useAggregatedData` is probably closer to this convention that `useCreateAggDataOnRes`.

eliascodes:
> So RE: what we spoken about on slack:

Alternatively, we could move all formatting code to within the data table component. So no formatting/rounding of numbers happens before the data is passed to the table, and same with the months. That is actually not a bad idea.

Issue #101, opened 2019-05-21
Changes to data manipulation


eliascodes:
As a result of review of #90 

### Coupling

* [x] Remove the `Total ${unit}` column from `logstoAggregatedData` (move it to `aggregatedToTableData`): this couples the aggregate data structure too closely to the design of the tables; it should also support the charts which won't have this data point.

* [x] Don't pass column headers to `logsToAggregateData`: `AggregateData` is an intermediate data structure that should be de-coupled from table-specific considerations. Instead just identify the fields to use as column headers (already done via `tableTypes`) and use those to accumulate the columns represented in the log data. Specific column headers/ordering can be specified in the next step `aggregatedToTableData`, and should replace empty or undefined keys with `0`.

### Composition

* [x] Compose the `useRequest` hook with the `useAggregateData`

eliascodes:
After taking a closer look at the code (after actually having to change it in #117), I have the following comments.

Related to my second point above, ultimately what we have is a process of _projecting_ a list of objects onto a single "object". In `logsToAggregatedData` we want to arrive at a point where all our log data is combined in a data structure that has, at least, _rows_ and _cells_.
```ts
type Cell = string | Duration;

interface Row extends Dictionary {}

type TableContent = Row<Cell>[];

interface Table {
  headers: string[];
  rows: TableContent;
}

interface Projection {
  <T>(rowId: string, colId: string, combinator: (a: T, b: T) => T, data: Dictionary<T>[]): TableContent;
}
```
We can additionally define functions that map over `Row`s and `Cell`s:
```ts
interface mapRow {
  (f: (c: Cell) => Cell, r: Row): Row;
}

interface mapTableContent {
  (f: (r: Row) => Row, rs: TableContent): TableContent;
}
```
These can be used to implement all the date string transformations and the totals calculation in `aggregateToTableData`, and also to pad the columns with default values (e.g. zeroes for columns with no data).

The point of this is that if we identify and extract the more primitive concepts, it becomes much easier to (1) see whats going on, (2) extend it and (3) re-use it.

The use of `Table` in the type names is mostly out of a lack of a better name for a labelled 2D grid, but obviously these shouldn't necessarily be conceptually bound to the "tables" we see in the actual pages (as may become clear once we start doing charts stuff).

astroash:
- what does `combinator` describe in the above?
- how does this differ from the deriving column headers from the available data as previously discussed?

eliascodes:
[...]

PR #106, opened 2019-05-22
🐛 fix abbreviated month bug on firefox

### LINKED ISSUE #105: Data table dates are "invalid"
Presumably as a result of #90

Found on `twine-dashboard-qa` and `twine-dashboard-staging`. This appears not to be a problem when I try to reproduce locally, suggesting an issue with the data itself?

Screenshot 2019-05-22 at 14 42 03

astroash:
- create ISO string format before creating moment/date object

fixes #105

### DIFF
--- a/dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
+++ b/dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
@@ -24,7 +24,6 @@ export const abbreviateDateString = (x: string): string => {
     ? `${numericalMonth}`
     : `0${numericalMonth}`;
   const isoDate = `${year}-${doubleDigitNumericalMonth}`;
-  console.log(isoDate);
   return moment(isoDate).format(Months.format.table);
 };
 
--- a/dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
+++ b/dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
@@ -18,7 +18,14 @@ export const isDateString = (x: any): boolean => {
 };
 
 export const abbreviateDateString = (x: string): string => {
-  return moment(new Date(x)).format(Months.format.table);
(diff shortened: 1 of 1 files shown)


dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
```diff
@@ -18,7 +18,14 @@ export const isDateString = (x: any): boolean => {
...
 
 export const abbreviateDateString = (x: string): string => {
-  return moment(new Date(x)).format(Months.format.table);
+  const [month, year] = x.split(' ');
+  const numericalMonth = Months.list.indexOf(month) + 1;
+  const doubleDigitNumericalMonth = numericalMonth > 9
+    ? `${numericalMonth}`
+    : `0${numericalMonth}`;
+  const isoDate = `${year}-${doubleDigitNumericalMonth}`;
+  console.log(isoDate);
```
eliascodes: 🔥

dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
```diff
@@ -18,7 +18,14 @@ export const isDateString = (x: any): boolean => {
 };
 
 export const abbreviateDateString = (x: string): string => {
-  return moment(new Date(x)).format(Months.format.table);
+  const [month, year] = x.split(' ');
```
eliascodes: Is this all necessary because the `MMM YY` string isn't parsed by the date constructor in Firefox?

astroash: At this point date is represented by `MMMM YYYY` eg January 2013. Moment only handles dates that are iso or rfc2822 format so it is passed to the date constructor instead (functionality that is going to be removed in future releases of `moment`).

This could be avoided if we use different month string formatting in the aggregation layer which we can look into as part of #101

Issue #98, opened 2019-05-21
📏 table sorting defects


astroash:
- [x] Sorting should default to alphabetical for duplicate values
> VOL-DASH-003-0003 - Sorting
Steps: View the table under Activity.
Sort on a col where the data is all the same for all volunteers.
Expected:  When sorting by a column (eg, Training/Teaching) where more than one Volunteer has same hours, the secondary order is alphabetical by Volunteer Name
Actual:   Order of names appear in an adhoc order

- [x] default sorting
> VOL-DASH-007-0005
Steps: View the table under Time- intial default view of VOL-DASH-007 
Expected: The table data and graph key/legend to be initially (default view) sorted by ‘Activity’ - alphabetical ascending A-Z.
Actual: Table sorted by alphabetical descending Z-A

- [x] csv sorting
> VOL-DASH-007-0009
Steps: View the table under Time- intial default view of VOL-DASH-007.  Observe the sort order (by Activity) and click download button
Expected: The download data sorted by ‘Activity’ - alphabetical ascending A-Z.
Actual: The sort order is random - not sorted by Activity (A-Z) as expected but also not sorted by any of the other columns

- [x] reverse alphabetical order
> "Steps: View the table under Activity.  Note that it is sorted by Total Hours.  Click the Volunteer name column heading to sort by Volunteer Name
Expected:  Names to be sorted alphabetically A-Z
Actual:  Names are sorted alphabetically Z-A
This also affects the Volunteer table as well as when you resort a text-data column eg. Activity (in the Time table)"

astroash:
csv sorting not resolved

PR #109, opened 2019-05-22
🐛 Match initial sorting with user stories

eliascodes:
### Changes
* Add `sortBy` to arguments of `aggregatedToTableData` so each table can specify initial sort column


astroash:
approved pending 108

astroash:
related #98

PR #117, opened 2019-05-24
🐛 Fix table sorting on unit switch

### LINKED ISSUE #96: 🐛 totals column sorting on unit change
> Sort by Volunteer name
Swap from hours to days near top right
Table now sorts by hours or days. 
Intentional? Rather than swapping display of hrs/days it also sorts by it automatically?

eliascodes:
Fixes #96 
Related to #101

### Changes
* Store table `title` and `sortBy` separately from other table props (`headers` and `rows`). 
  * The former is entirely static and doesn't need to be in state
  * The latter is a presentational concern not related to the actual content of the table/data
* Do not calculate row totals in `logsToAggregateData`, but in `aggregateToTableData` instead (see #101)
* Minor simplifications and type tightening in `logsToAggregateData`
* Add more object utilities to `twine-util`.


dashboard-app/src/dashboard/ByActivity/index.tsx
```diff
@@ -29,13 +28,16 @@ const Container = styled(Grid)`
...
+
 const ByActivity: FunctionComponent<RouteComponentProps> = (props) => {
   const [unit, setUnit] = useState(DurationUnitEnum.HOURS);
+  const [sortBy, setSortBy] = useState(`Total ${unit}`);
   const [activities, setActivities] = useState([]);
   const [volunteers, setVolunteers] = useState();
   const [fromDate, setFromDate] = useState(Months.defaultFrom());
   const [toDate, setToDate] = useState(Months.defaultTo());
-  const [tableProps, setTableProps] = useState<DataTableProps | null>();
+  const [tableProps, setTableProps] = useState<Pick<DataTableProps, 'headers' | 'rows'> | null>();
```
astroash: This variable no longer represents table props. Can this be renamed to `tableRows` || `tableData` || somethings similar?

dashboard-app/src/util/dataManipulation/logsToAggregatedData.ts
```diff
@@ -1,56 +1,33 @@
...
-  }), rows));
-
-
 interface Params {
   logs: any;
   columnHeaders: string[];
-  unit: DurationUnitEnum;
   tableType: TableTypeItem;
-  volunteers?: any;
+  volunteers?: { id: string, name: string }[];
```
astroash: Can we add a `//TODO` to replace this with api type?

dashboard-app/src/util/dataManipulation/aggregatedToCsv.ts
```diff
@@ -1,15 +1,18 @@
...
 import { AggregatedData } from './logsToAggregatedData';
+import { aggregatedToTableData, abbreviateDateString } from './aggregatedToTableData';
+import { DurationUnitEnum } from '../../types';
+import { Objects } from 'twine-util';
+
+
+// tslint:disable-next-line: max-line-length
+export const aggregatedToCsv = async (data: AggregatedData, unit: DurationUnitEnum): Promise<string> =>
+  new Promise((resolve, reject) => {
+    const tableData = aggregatedToTableData({ data, unit });
```
astroash: Can we manipulate this data with the functionality that it needs? Seems  counter productive to separate out functionality to then reuse `aggregatedToTableData`

eliascodes: I was going for minimal diff, but if you're happy for me to do a bit of re-organisation then sure.

eliascodes: Actually turns out I only needed one of the exported util functions, didn't re-organise anything, lol

dashboard-app/src/util/dataManipulation/aggregatedToTableData.ts
```diff
@@ -47,12 +50,37 @@ const abbreviateMonths = evolve({
...
+      return round(Duration.toWorkingDays(duration));
+
+    case DurationUnitEnum.HOURS:
+      return round(Duration.toHours(duration));
+  }
+};
+
+const add = (a: number, b: any) => a + (isNaN(b) ? 0 : Number(b));
+
+export const calculateTotalsUsing = (unit: DurationUnitEnum) => evolve({
```
astroash: Can we pull this function out of this file? it's being used by csv manipulation & `aggregatedToTableData`

eliascodes: Do you mean the `Using`?

astroash: Yup

astroash: I don't think the functionality is clear from the naming. This function also does more than calculate, it also adds these values to the data structure. Approving as keen to get this in, but would be up for a rename in future.

PR #125, opened 2019-05-29
🐛 fix nav-bar page order

eliascodes:
and add test to check order


eliascodes:
> Is this depending on object order to for the navbar?

Yes. Since ES2015, object property order is part of the spec. Though broadly I think I agree with specifying the order explicitly anyway.

> Can we be explicit with the ordering

I used a separate function rather than changing the type because `order` is not an inherent property of a "page". Pages exist as nodes in a graph rather than as elements of an ordered list. The notion of "ordering" is only relevant to the navigation bar, so it gets its own function.

Issue #123, opened 2019-05-28
🐛 max connections issues


astroash:
When downloading csv files on the visitor app server gives max connection error 

Investigate why this is happening

astroash:
Posting a bit of background research:
- db max connections set to `500` by default by heroku
- role connection limit set to `20` by heroku.
can be checked with query:
```
SELECT rolname, rolconnlimit
FROM pg_roles
WHERE rolname = [db_user_name];
```
- it seems common practice to set this value to `-1`, but i imagine has set this due to tier restrictions and wouldn't like us messing around with values. `pg` max pool is also set to `20` so this shouldn't be an issue (max connections per user reference)[https://confluence.atlassian.com/jirakb/fatal-too-many-connections-error-when-using-postgresql-database-due-to-user-specific-connection-limit-238617669.html]
- do pg & postgres use the same counting system (ie is one zero based?)

eliascodes:
I think the problem is on this line:
https://github.com/TwinePlatform/twine-monolith/blob/9034631002f16f64feb090f31059fc6ba9026f48/api/src/models/visitor.ts#L188

Using `Promise.all` sets everything off in parallel, causing Knex to grab as many connections as it can and using up the limit. I don't know why the max connection limit option doesn't help sort this out tbh.

I want to try two things:
1. Speculative -- Reduce the connection limit to 19. It may be that Heroku actually means `<20` rather than `<= 20`.
2. Almost definitely will work -- Use `Promises.series` from `twine-util` instead of `Promise.all`

When trying option 1 I also want to make this an option that can be changed by environment variable.

astroash:
> Speculative -- Reduce the connection limit to 19. It may be that Heroku actually means <20 rather than <= 20.

You could test this locally by setting the `rolconnlimit` to 20 of the user you connect to the db with

eliascodes:
I can't update `pg_roles`, it's a read-only view, not a table.

astroash:
`ALTER ROLE [db_user] CONNECTION LIMIT -1;`

eliascodes:
tbh I haven't been able to reproduce this locally anyway. I think our demo data sets are too small

eliascodes:
On production, I can reproduce this bug on the test account which has ~260 visits.

On staging, I tried to verify this fix by:
- Inserting visits into the demo account so there are ~250 total
- Exporting visits data to CSV

The bug did not appear (i.e. was able to successfully download data).

[...]

PR #137, opened 2019-05-30
🐛 limit concurrent queries by replacing Promise.all with Promises.series

eliascodes:
Related to #136
Related to #123 

Only merge this if #136 proves insufficient in resolving #123


eliascodes:
Yeah so #136 doesn't work, fixing tests on this now :)

Issue #130, opened 2019-05-30
📊 Chart js manipulation


astroash:
### Chart Designs:
<img width="680" alt="Screenshot 2019-05-30 at 11 24 10" src="https://user-images.githubusercontent.com/14337958/58623084-6d968b80-82cd-11e9-9420-110cd5a16e86.png">

### Out of the box chart:
<img width="915" alt="Screenshot 2019-05-30 at 11 02 41" src="https://user-images.githubusercontent.com/14337958/58623086-6ff8e580-82cd-11e9-84cc-d06e462e4f7a.png">

### Changes:
Chart:
- [ ] border radius on bars
- [ ] total on top of bars
- [ ] remove number scale
- [ ] text direction of x axis lables

Legend:
- [ ] title
- [ ] separate element with box shadow

astroash:
**Current chart status:**

<img width="997" alt="Screenshot 2019-05-31 at 17 34 06" src="https://user-images.githubusercontent.com/14337958/58717152-de6f9d80-83ca-11e9-8619-97072f493650.png">

🐛 known bug for users with the same name interfering with graph filtering

eliascodes:
## Review missing features from user stories

### 006B
* Title should include information about the filter (e.g. "Jun 2018 - Jan 2019")
* "No data available" view
* Hide "download" button when chart is displayed

### 007B
* Title should be "Volunteer Activity per month" not "Volunteer Activity over months"
* Title should include information about the filter (e.g. "Jun 2018 - Jan 2019")
* "No data available" view
* Hide "download" button when chart is displayed

PR #139, opened 2019-06-03
📊 display customised graphs in tabs

astroash:
**Changes:**
- remove steps & lines
- add totals
- implement different date formatting for charts/tables/csv
- refactor chart config
- breakout tabs component
- refactor errors
- sneaky white bg to toggle

related #130


dashboard-app/src/components/StackedBarChart/util.ts
```diff
@@ -0,0 +1,71 @@
+import { round } from '../../dashboard/dataManipulation/util';
```
eliascodes: As a general point, this is a little strange IMO. Dependencies should not, where reasonable, go down the abstraction/generality tree.

To be a little less abstract: components is a directory full of (relatively) generic components, not specific to the requirements of a given feature. Whereas dashboard is concerned with assembling the more generic components into something that satisfies the specific requirements of the features it is trying to implement.

If I delete everything from dashboard, nothing in components (or anywhere else in the app, really, except App.tsx and navigation) should break.

So I think either both the dashboard and the components should rely on some common rounding function in a top level utils, or they should both import from twine-util and specify their own rounding functions (minor preference for the latter because the functionality (rounding) is so trivial).

dashboard-app/src/components/ErrorParagraph.tsx
```diff
@@ -1,14 +1,37 @@
...
-export const displayErrors = (errors: Dictionary<string>) =>
-  (<ErrorParagraph>{toPairs(errors).map((x: any) => x.join(': ')) || ''}</ErrorParagraph>);
+
+/*
+ * Component
+ */
+const Errors: FunctionComponent<Props> = (props) =>
+  (<Row center="xs">
+    <Col>
+      <ErrorParagraph>{toPairs(props.errors).map((x: any) => x.join(': ')) || ''}</ErrorParagraph>
```
eliascodes: There's no reason for this to be `any`, TS can infer the type

dashboard-app/src/dashboard/ByTime/TimeTabs.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * Types
+ */
+
+interface Props {
+  aggData: AggregatedData;
+  unit: DurationUnitEnum;
+  tableData: TableData;
+  sortBy: number;
+  onChangeSortBy: (x: string) => void;
+  TITLE: string;
```
eliascodes: why is this in caps?

dashboard-app/src/dashboard/ByVolunteer/VolunteerTabs.tsx
```diff
@@ -0,0 +1,63 @@
...
+`;
+
+/*
+ * Component
+ */
+const VolunteerTabs: FunctionComponent<Props> = (props) => {
+  const { aggData, unit, tableData, sortBy, onChangeSortBy, TITLE } = props;
+
+  return(
+  <TabGroup titles={['Chart', 'Table']}>
```
eliascodes: Indentation is all over the place in these files, we should fix the linting at some point in the future.

Issue #133, opened 2019-05-30
Hooks refactor


eliascodes:
In the process of working on #129 it became clear that the hooks could do with a bit of work; at the moment they're a little too low-level and require a bit too much boilerplate.

### General
Looking at a lot of the [custom utility hook libraries](https://nikgraf.github.io/react-hooks/), the idiomatic approach seems to be to return values out of the hook rather than accept callbacks as arguments. I think we should use this approach.

### Hooks
Ordered in terms of dependency and my opinion of usefulness:
* [`useAsync`]()
* [`useRequest`]()
* [`useBatchRequest`]()
* [`useAggregatedDataByActivity`]()
* [`useAggregatedDataByTime`]()
* [`useAggregatedDataByVolunteer`]()
* [`useInteger`]()
* [`useToggle`]()
* [`useDate`]()

### useAsync
```ts
const [state, run] = useAsync(async () => {});

state.loading; // bool
state.error; // Error | null
state.result; // {}
```

### useRequest
Uses: `useAsync`
```ts
const [state, run] = useRequest({ 
  call: async () => {},
  params: {},
  updateOn: [],
  immediate: false,
});

state.loading; // bool
state.error; // Error | null
state.result; // {}
```

### useBatchRequest
Uses: `useRequest`
```ts
const [states, run] = useBatchRequest([
  {
    call: async () => {},
    params: {},
    updateOn: [],
    immediate: false,
  },
  // ...
]);

states.all[0] // { loading, results, error }[]
states.loading // boolean
states.results // {}[]
states.errors // (Error | null)[]
```

### useAggregatedDataByActivity
Uses `useBatchRequest`
```ts
const { data, loading, error } = useAggregatedDataByActivity({
  updateOn: [],
  from: Date,
  to: Date,
});
```

### useAggregatedDataByTime
Uses `useBatchRequest`
```ts
const { data, loading, error } = useAggregatedDataByActivity({
  updateOn: [],
  from: Date,
  to: Date,
});
```

### useAggregatedDataByVolunteer
Uses `useBatchRequest`
```ts
const { data, loading, error } = useAggregatedDataByActivity({
  updateOn: [],
  from: Date,
  to: Date,
});
```

### useInteger
```ts
const [num, setNum, inc, dec] = useInteger({ init: 0, min: -1, max: 10 });
```

### useToggle
```ts
const [on, onToggle] = useToggle(false); // simple boolean toggle

[...]

PR #140, opened 2019-06-03
♻️ Hooks refactor

eliascodes:
Related to #133 
Related to #101 

* Upgrade Typescript to 3.5.1
* Adds `react-use` and required peer-dependencies (`keyboardjs` and `rebound`)
* Change `useRequest` to use `useAsyncFn` from `react-use`
* Add `useBatchRequest` custom hook to make multiple HTTP requests in parallel
* Add `useAggregatedDataBy<X>` custom hooks to form aggregated data required by each data view
* Add meaningful tests on page components
* Mock out charts on tests (they crash when running via JSDOM even with `canvas` installed)


dashboard-app/src/api/index.ts
```diff
@@ -35,12 +35,19 @@ export const CbAdmins = {
 };
 
 export const CommunityBusinesses = {
+  configs: {
```
astroash: What's the benefit of separating out the method and url?

dashboard-app/src/dashboard/ByActivity/index.tsx
```diff
@@ -132,25 +102,30 @@ const ByActivity: FunctionComponent<RouteComponentProps> = (props) => {
...
-                initialOrder="desc"
-                onChangeSortBy={onChangeSortBy}
-                showTotals
-              />
-            )
-          }
-        </Col>
-      </Row>)
+      {
+        loading
```
astroash: we should remove the loading spinner on subsequent requests, but we can do this in another PR.

dashboard-app/src/dashboard/ByTime/index.tsx
```diff
@@ -84,10 +67,10 @@ const ByTime: FunctionComponent<RouteComponentProps> = (props) => {
   }, [tableData]);
 
   const downloadAsCsv = useCallback(() => {
-    downloadCsv({ aggData, fromDate, toDate, setErrors, fileName: 'by_activity', unit });
-  }, [aggData, fromDate, toDate, unit]);
+    downloadCsv({ aggData: data, fromDate, toDate, setErrors, fileName: 'by_activity', unit });
```
astroash: I think we should change all data manipulation functions to take a `data` arg rather than `aggData`

dashboard-app/src/dashboard/ByVolunteer/__tests__/index.test.ts
```diff
@@ -14,16 +19,73 @@ describe('By Activity Page', () => {
...
+        tools.getByText('Wilma', { exact: true }),
+      ]);
 
-    mock.onGet('/community-businesses/me/volunteer-logs').reply(200, { result: [] });
-    mock.onGet('/community-businesses/me/volunteers').reply(200, { result: [] });
+      await waitForElement(() => tools.getByText('Totals', { exact: true }));
 
-    const tools = renderWithHistory(ByVolunteer);
+      const rows = await waitForElement(() => tools.getAllByTestId('data-table-row'));
+      // console.log(rows.map((row) => row.textContent));
```
astroash: 🔥 remove all console.logs

dashboard-app/src/dashboard/hooks/useAggregatedDataByActivity.ts
```diff
@@ -0,0 +1,74 @@
+import { DependencyList, useEffect, useState } from 'react';
```
astroash: I think these hook names should be present tense

`useAggregateDataByActivity`

dashboard-app/package.json
```diff
@@ -57,9 +58,11 @@
     "react-router-dom": "^4.3.1",
     "react-scripts": "2.1.8",
     "react-spinners": "^0.5.3",
+    "react-use": "^9.1.2",
+    "rebound": "^0.1.0",
     "styled-components": "^4.1.3",
     "twine-util": "file:../lib/twine-util",
-    "typescript": "^3.3.4000"
+    "typescript": "^3.5.1"
```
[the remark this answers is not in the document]
eliascodes: I didn't see any issues related to styled components, but a couple of new build issues did arise in the data manipulation code.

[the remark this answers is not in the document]
astroash: the SC type issue was specifically a bug about compile time with vscode. Probs been fixed by now :)

dashboard-app/src/api/index.ts
eliascodes: It's more a question of having a static declaration of the requests. 

If we only have functions that make the requests, then `useBatchRequest` needs to take `<T>(...a: any) => Promise<T>` as arguments, at which point it becomes less `useBatchRequest` and more `useBatchAsync` since there's no way of restricting those functions to only making API calls (except by manual inspection).

If we're gonna use hooks to do all API calls (eventually), then I imagine we can (also eventually) get rid of the function calls and just keep the static declarations.

eliascodes: ...unless we make the it accept `(...a: any) => Promise<AxiosResponse>`. But then there's still no guarantee that the function isn't doing arbitrary side effects.

dashboard-app/src/components/StackedBarChart/index.tsx
```diff
@@ -15,7 +15,6 @@ import { ColoursEnum } from '../../styles/design_system';
 /*
  * Types
  */
-
```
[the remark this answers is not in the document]
eliascodes: Ok, but can you fix the formatting then?

PR #182, opened 2019-06-20
📊 customise tooltip

### LINKED ISSUE #166: Custom tooltip changes
* [x] Positioning:

> Sometimes the tooltips are not as clear what part of the bar they are referring to as the they are aligned to the top. This is especially an issue when there are small bar sections. Is there a setting to align them to the centre of the section of the bar as per the designs?

* [x] Contrast ratio:

> Poor contrast ratio where volunteer colours are against the dark background of the tooltip. Style tooltip as per designs if possible.

* [x] Units:

> The tooltip does not show the unit of time after the value for that name. Add "hrs" or "Days" after the value. 

* [ ] ~~Date formatting~~ moved date as not in designs

> Remove the comma after DDD and before YYYY

astroash:
**Changes:**
- change colour to match designs
- position in center of bar
- set title of tt to total with unit (hrs/days)
- set description of tt to categeorgy w/o total

fixes #166

### DIFF
--- a/dashboard-app/src/components/StackedBarChart/Chart.tsx
+++ b/dashboard-app/src/components/StackedBarChart/Chart.tsx
@@ -11,7 +11,6 @@ import _Card from '../Card';
 import { H3 } from '../Headings';
 import { ColoursEnum } from '../../styles/design_system';
 import { FullWidthTextBox } from '../FullWidthTextBox';
-import { DurationUnitEnum } from '../../types';
 
 
 /*
--- a/dashboard-app/src/components/StackedBarChart/Chart.tsx
+++ b/dashboard-app/src/components/StackedBarChart/Chart.tsx
@@ -11,6 +11,7 @@ import _Card from '../Card';
 import { H3 } from '../Headings';
 import { ColoursEnum } from '../../styles/design_system';
 import { FullWidthTextBox } from '../FullWidthTextBox';
+import { DurationUnitEnum } from '../../types';
 
 
 /*
@@ -23,6 +24,7 @@ interface Props {
   title: string;
   noActiveLegendText: string;
   isAllLegendDataInactive: boolean;
+  unit: DurationUnitEnum;
 }
 
 type HidableTextBoxProp = {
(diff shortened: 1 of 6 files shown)


dashboard-app/src/components/StackedBarChart/Chart.tsx
```diff
@@ -65,14 +67,15 @@ const Chart: FunctionComponent<Props> = (props) => {
     title,
     isAllLegendDataInactive,
     noActiveLegendText,
+    unit,
```
eliascodes: I don't know whether #183 was in part motivated by this kind of change, but my specific suggestion here would be that, it is actually quite reasonable for a "Chart" to be told what units the quantities it is displaying are measured in. 

Depending on your perspective about how generic this component should be, it is perhaps less reasonable for it to be given specific knowledge about hours/days (as is happening in `getStackedBarChartOptions`).

My suggestion would be that rather than `unit` being passed down, instead we pass down `xAxisUnit: string` and `yAxisUnit: string`. This is more natural for a chart component, more generic, and allows us to label axes with months if we wish to do so later on.

So we would pass down `{ yAxisUnit: 'hrs' }` or `{ yAxisUnit: 'days' }` from the parent. I haven't yet figured whether either prop should be optional.

dashboard-app/src/components/StackedBarChart/utils/chartjsUtils.ts
```diff
@@ -25,47 +27,72 @@ export const totalizer = {
...
-    }],
-    yAxes: [{
-      scaleLabel: {
-        display: true,
-        labelString: yAxisTitle,
+
+export const getStackedGraphOptions = (
+  xAxisTitle: string,
+  yAxisTitle: string,
+  unit: DurationUnitEnum): any => ({
```
eliascodes: Does this need to be `any`?

astroash: seemed to be something in my package-lock causing me to need this 🤷🏽‍♀️. Gone now...

dashboard-app/src/components/StackedBarChart/Chart.tsx
astroash: `getStackedBarChartOptions` is already very tied to the particular type of graph we want to create. If we do end up displaying multiple types of the graph then we should split out the options so that we are mutating `Chart` with our default config, then generate specific chart options via a funcs similar to this.

PR #180, opened 2019-06-20
🐛 Add specifc CSV error msg when data is empty

eliascodes:
#### Changes
- Check if download failed before data is empty, and display message if so
- Add test on this case

Fixes #178

PR #184, opened 2019-06-20
🐛 Error msg persistence

### LINKED ISSUE #177: 🐛 Persistent csv error message
>Steps: Browse to the Activity table. 
Change the date filter to a range where there is no data - eg. 19/06/2019-19/06/2019 
Click the Download Csv button
Observe the error message - Download: There was a problem downloading your data
Change the filter to another date range WITH data.
Actual: Error message persists and does not disappear
Expected: Error message disappears

eliascodes:
Merge #180 first

### Changes
- Remove 'Download' error message everytime `error`, `fromDate` or `toDate` changes
- Extract `setError` operation to top-level component to avoid passing state setters directly to children

Fixes #177 

NOTE: There is perhaps a concept her calling for a hook called something like `useStateUnless` which sets something in state unless a given condition is satisfied, at which point that value is unset. The API didn't immediately come to me though, so I figured I'd leave it. Something to think about maybe though.


astroash:
@eliascodes see #185 for a potentially simpler fix

astroash:
superseded by #185

PR #185, opened 2019-06-21
🐛 use custom hook to ensure error messages are cleared

### LINKED ISSUE #177: 🐛 Persistent csv error message

### LINKED ISSUE #187: Un-expressed/un-enforced dependency between chart/legend data
Related to #186 

See comment:

> I'm merging for expediency, but this seems to indicate a dependency between the legend data and chart data that isn't enforced. They should be derived from the same source in the same way/same order, so that this otherwise obtuse .reverse isn't necessary.

astroash:
**Changes:**
- clear all errors on successful response (we don't have any inputs validation so no need for errors to persist)
- set errors on response error
- create a hook to do all the work
- move page specific hooks to page folder

fixes #177
fixes #187 
alternative to #184


dashboard-app/src/hooks/useErrorsEffect.ts
```diff
@@ -0,0 +1,21 @@
+import { Dispatch, SetStateAction, useEffect } from 'react';
+import { AxiosError } from 'axios';
+import { AggregatedData } from '../dashboard/dataManipulation/logsToAggregatedData';
+import { Dictionary } from 'ramda';
+
+export const useErrorsEffect = (
+  error: AxiosError | undefined,
```
eliascodes: `error?:` and `data?:` instead of `| undefined`

dashboard-app/src/dashboard/ByActivity/useAggregateDataByActivity.ts
```diff
@@ -0,0 +1,79 @@
+import { DependencyList, useEffect, useState } from 'react';
```
eliascodes: Why is this showing up as a new file? Did you copy it instead of moving it?

dashboard-app/src/hooks/useErrorsEffect.ts
```diff
@@ -0,0 +1,21 @@
+import { Dispatch, SetStateAction, useEffect } from 'react';
+import { AxiosError } from 'axios';
+import { AggregatedData } from '../dashboard/dataManipulation/logsToAggregatedData';
+import { Dictionary } from 'ramda';
+
+export const useErrorsEffect = (
```
eliascodes: There is some core semantic being expressed here that has nothing particularly to do with either errors or aggregate data.

It feels like something like this, which is something I was trying to grasp at in my note in #184:
```ts
const useResettableState = <T>(init: T, updateOn: any[], resetOn: any[]) => {
	// ...
};

// Usage
const [errors, setErrors] = useResettableState({}, [error], [data]);
```

dashboard-app/src/hooks/useErrorsEffect.ts
astroash: `?` makes it an optional argument, which messes with arg order

dashboard-app/src/dashboard/ByActivity/useAggregateDataByActivity.ts
astroash: Nope, moved it.

dashboard-app/src/hooks/useErrorsEffect.ts
astroash: To get this working with the current implementation, we would also need to supply a callback to describe how we are updating state. I think it would be simpler if we removed `onUpdate`, and handled this separately:

```js
const useResettableState = <T>(init: T, resetOn: any[]) => {
	// ...
};
const [errors, setErrors] = useResettableState({}, [data]);
```

eliascodes: Yeah good point. I suspect if I had tried to fill out `// ...` in my example I'd have realised that, lol.

However, we could still maybe keep your error hook, and just compose it with a `useResettableState` hook. 

So in the end this might look like:
```ts
const useAxiosErrors(resetOn: any[]) => {
  const [errors, setErrors] = useResettableState<Dictionary<string>>({}, resetOn);

  const setState = useCallback((error: AxiosError, key: string) => {
    setErrors(assoc(key, error.message))
  }, []);

  return [errors, setState];
}
```

dashboard-app/src/dashboard/ByActivity/useAggregateDataByActivity.ts
eliascodes: No you didn't:

Screenshot 2019-06-24 at 11 15 21

dashboard-app/src/hooks/useAxiosErrors.ts
```diff
@@ -0,0 +1,21 @@
...
+  const [errors, setErrors] = useResettableState<Dictionary<string>>({}, [data]);
+
+  // set errors on response error
+  useEffect(() => {
+    if (error) {
+      setErrors({ data: error.message });
+    }
+  }, [error]);
+
+  return [errors, setErrors] as [Dictionary<string>, Dispatch<SetStateAction<Dictionary<string>>>];
```
eliascodes: Not sure this interface makes sense. You're allowing the caller to set the `errors` state directly via returning `setErrors`, but also updating it automatically everytime `error: AxiosError` changes.

I think it should probably look either like my example in my previous comment, or just don't return `setErrors`. If this hook is only for a specific pattern, then maybe the latter is fine.

astroash: I don't understand your implementation above. What would it look like in the page element?

eliascodes: Well you'd have to lift the `useEffect` call out into the parent, so probably your version is better so we can avoid that. I guess just don't return `setErrors`, unless you were thinking of a particular use case that I can't see?

dashboard-app/src/dashboard/ByActivity/useAggregateDataByActivity.ts
astroash: i blame git...

dashboard-app/src/hooks/useAxiosErrors.ts
astroash: That won't work. `setErrors` needs to be passed to any function on the page that is doing error handling. 

I think my imagining of `useAxiosErrors` is different to yours. I was imagining it a fancy custom useState hook that has bespoke special useEffect magic. A bit like a higher order component but for hooks, the `setState` still has it's standard state setting functionality but ✨ *_upgraded_* ✨

Here's what it can do:
- able to set initial state
- able to reset state on change of 
- able to be updated as an when we want

In app use case:
- errors set to empty object on initial load
- if any request errors it updates the error from the request
- if any manipulation errors (table/graph/csv parsing) happen it is updated
- if any new request is made it wipes all errors
	- this is because all manipulation errors are based off of the request data. If new data is requested, it is assumed that the previous error is no longer relevant.

3rd bullet point is why we need to pass out `setErrors` so that data manipulation functions can update errors accordingly

eliascodes: Ah ok, so we want this to be used not only for Axios errors then. Maybe the name I suggested is misleading, do you have a better idea for one?

astroash: -`useErrors`
-`setAndUseErrors`
-`setUseResetErrors`

eliascodes: Lets go with `useErrors` for now then.

astroash:
@eliascodes, struggling to get tests working on this. Tempted to merge as is for QA, and add tests separately

dashboard-app/src/dashboard/dataManipulation/aggregatedToGraphData.ts
```diff
@@ -25,7 +25,11 @@ export const aggregatedToStackedGraph = (data: AggregatedData, unit: DurationUni
         data: labels.map((y) => numericData[y]),
       };
     })
-    // display stacks in same colour order as legend
+    // Reverse data-set order because Chart.js iterates
```
astroash: sneaky comment 🙃

Issue #183, opened 2019-06-20
✨ State management system


astroash:
State is currently being passed down multiple levels of components. This creates complexities when refactoring, and involves types being imported into every components the state is passed to via props. It would be beneficial to introduce some type of state management such as redux or reacts context api.

PR #205, opened 2019-06-30
♻️ context demo

astroash:
relates #183 

### Changes
- use context for app wide state for unit
- update toggle and util bar to support this
- add default value not to break tests

### Testing Requirements
- [x] **Dashboard App**
  - change unit on time page
  - observe new unit persistent on all pages

### Release Requirements
- na - release when ready

### Manual Deployment
- dashboard

### DIFF
--- a/dashboard-app/src/App.tsx
+++ b/dashboard-app/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { createContext, useState } from 'react';
 import styled from 'styled-components';
 import { Route, Switch, BrowserRouter, withRouter } from 'react-router-dom';
 
@@ -16,6 +16,7 @@ import ErrorPage from './features/Error';
 import Navbar from './features/navigation/Navbar';
 import Footer from './lib/ui/components/Footer';
 import { ColoursEnum, Fonts } from './lib/ui/design_system';
+import { DurationUnitEnum } from './types';
 
 /*
  * Styles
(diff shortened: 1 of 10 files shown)


astroash:
To be revisited after tiles implementation

eliascodes:
I know your main concern in doing this was component state and minimising prop-drilling, but I also think this gives us a chance to look at where our data is loaded.

I don't think we need to change how we're loading data now, but at least thinking about it will hopefully prevent us from implementing something now that will be a hurdle in a few months.

So I think it's useful to draw out a hierarchy of components and the data they require. I think what we have at the moment looks more or less like:

```
App
|
|- Switch
|  |- Dashboard
|  |  |~ `VolunteerLog[]` (12 months)
|  |  |~ `User[]` (All)
|  |  |~ `VolunteerActivity[]` (All)
|  |- Time
|  |  |~ `VolunteerLog[]` (Variable)
|  |  |~ `VolunteerActivity[]` (All)
|  |- Activity
|  |  |~ `VolunteerLog[]` (Variable)
|  |  |~ `VolunteerActivity[]` (All)
|  |  |~ `User[]` (All)
|  |- Volunteer
|  |  |~ `VolunteerLog[]` (Variable)
|  |  |~ `User[]` (All)
```
Key: 
`|-` == child component
`|~` == data required by parent component

So this suggests that `User[]` and `VolunteerActivity[]` are common to all pages and could potentially live in a top-level context. Managing `VolunteerLog[]` in a top level context is also quite possible despite varying time windows in each page, but is probably not worth the effort at this stage.

Side note: the fact that we're loading in the data using a hook means we've de-coupled the data itself from the method that delivers the data. This means that if in future we fetch data from context rather than the API, we only need to change the hook!

We can do a similar thing for component state (excluding state fragments that just hold derivatives of API data):
```
App
|
|- Switch
|  |- Dashboard
|  |- Time
|  |  |~ `unit`
|  |  |~ `sortBy`
|  |  |~ `fromDate`
|  |  |~ `toDate`
|  |  |~ `legendData`
|  |- Activity
|  |  |~ `unit`
|  |  |~ `sortBy`
|  |  |~ `fromDate`
|  |  |~ `toDate`
|  |- Volunteer
|  |  |~ `unit`
|  |  |~ `sortBy`
|  |  |~ `fromDate`
|  |  |~ `toDate`
|  |  |~ `legendData`
```
Key: 
`|-` == child component
`|~` == state fragment used by parent component

So really `unit` is the only thing that can be pulled up to the top level because it's possible values are the same across all pages (which is what you've already done in this PR).

However, we don't need to only have top-level context. We can also position separate context elements lower down the hierarchy, for example:

```
App
|
|- Switch
|  |- ContextOne
|  |  |~ `unit`
|  |  |- Dashboard
|  |  |  |- Time
|  |  |  |  |- ContextTwo
|  |  |  |  |  |~ `sortBy`
|  |  |  |  |  |~ `fromDate`
|  |  |  |  |  |~ `toDate`
|  |  |  |  |  |~ `legendData`
|  |  |  |- Activity
|  |  |  |- Volunteer
```

This is one option. The other option is more redux-like: you just have one top-level context, and namespace state fragments that are only relevant to parts of the component tree:

```
{
  timePage: {
    fromDate: ...
  },
  activityPage: ...
  ...
}
```

We can also see something else: There are a number of state fragments that are common to the 3 (soon to be 4) data views, which could be bundled up using a hook. That way, as I mentioned above, if we decide that any of those values should come from context in the future, it will only require us to change the hook.

astroash:
> User[] and VolunteerActivity[]

Agree that we can create our own state management to keep this data at top level. This can either be included in a top level context or held independently.

> However, we don't need to only have top-level context.

As mentioned IRL context doesn't work as outlined in your diagram. Specific contexts need to be connected to, so we can't pass down `legendData` as this is a different value on each page. We can only add values to context that have the same name and value across pages.

Rebased this PR to work with master. Let me know if you have any specific comments re the setup of context here. Let's move future architecture discussions to another PR so they're not lost.

PR #209, opened 2019-07-01
🐛 prevent 500 error on login when user has no password

### LINKED ISSUE #88: 500 exception when user without pwd tries to login
https://github.com/TwinePlatform/twine-monolith/blob/d178be418de1788a9bf1c089a3c762de6536cfcb/api/src/api/v1/users/login.ts#L41

If user model doesn't have a password, the `compare` function throws an exception. This should in fact be `4xx` series error.

eliascodes:
Moving the role check before the payload check prevents this error.

Fixes #88

### DIFF
--- a/api/src/api/v1/users/__tests__/login.test.integration.ts
+++ b/api/src/api/v1/users/__tests__/login.test.integration.ts
@@ -170,13 +170,18 @@ describe('POST /users/login', () => {
       url: '/v1/users/login',
       payload: {
         restrict: ['VOLUNTEER'],
-        email: '1498@aperturescience.com',
+        email: '1498@aperturescience.com', // This is VISITOR!
         password: 'CakeisaLi3!',
       },
     });
 
     expect(res.statusCode).toBe(403);
     expect(typeof res.headers['set-cookie']).toBe('undefined');
-    expect(res.result).toEqual({ error: expect.objectContaining({ statusCode: 403 }) });
+    expect(res.result).toEqual({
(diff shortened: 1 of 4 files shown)


api/src/api/v1/users/__tests__/login.test.integration.ts
```diff
@@ -163,4 +163,20 @@ describe('POST /users/login', () => {
...
     expect(res.result).toEqual({ error: expect.objectContaining({ statusCode: 400 }) });
   });
+
+  test(':: unsuccessful login | type body | mismatched role', async () => {
+    const res = await server.inject({
+      method: 'POST',
+      url: '/v1/users/login',
+      payload: {
+        restrict: ['VOLUNTEER'],
+        email: '1498@aperturescience.com',
```
astroash: can you add a comment for which user types this is

api/src/api/v1/users/__tests__/login.test.integration.ts
```diff
@@ -163,4 +163,20 @@ describe('POST /users/login', () => {
...
+        password: 'CakeisaLi3!',
+      },
+    });
+
+    expect(res.statusCode).toBe(403);
+    expect(typeof res.headers['set-cookie']).toBe('undefined');
+    expect(res.result).toEqual({ error: expect.objectContaining({ statusCode: 403 }) });
```
astroash: Can we test the error message here too

Issue #67, opened 2019-05-15
Disallow CB_ADMIN users from creating logs for themselves


eliascodes:
There is currently nothing in the API preventing CB_ADMIN users from adding logs against their own account. This shouldn't be allowed.

https://github.com/TwinePlatform/twine-monolith/blob/6839343109d72c1ed71057923aceefec92286cd8/api/src/api/v1/community_businesses/volunteer_logs.ts#L246-L268

eliascodes:
fixed by #210

PR #210, opened 2019-07-01
🐛 disallow CB_ADMIN from adding own logs

### LINKED ISSUE #61: tables columns should be fixed width

eliascodes:
fixes #61 

Just add a role check if `userId === 'me'`, allowing only `VOLUNTEER` and `VOLUNTEER_ADMIN` users.


eliascodes:
This actually fixes #67 -- oops

PR #213, opened 2019-07-02
🔬 allow onChangeSortBy to return default sort order to address QA defect

eliascodes:
related to #98 

Allow `onChangeSortBy` to return default sort order.


astroash:
Not working when testing locally

steps:
1. go to time table
2. click sort by month
3. click "activity"
4. 👀 observe activity column is sorting z-a

eliascodes:
I only fixed the activities table. Are we interpreting this defect as relating to _all_ text based fields?

astroash:
Yeah, I think it means that all text columns should default to a-z sorting

eliascodes:
> I think it is meant to be by total hours

Nope, see the user story

Issue #192, opened 2019-06-24
📊 Create dashboard widgets


astroash:

astroash:
user stories being updated

eliascodes:
Suggested component interface:
```ts
type TileProps = {
  topText: string
  dataPointOne: string[]
  dataPointOneLabel: string
  dataPointTwo: number | string[]
  dataPointTwoLabel: string
};
```
Where `dataPointOne` is the list of activities or months (US010A/B) or empty (US010C), `dataPointOneLabel` is the text just above that, `dataPointTwo` is the number of hours (US010A/B) or the list of volunteer names (US010C), and `dataPointTwoLabel` is the text just below the yellow line.

The naming convention `dataPoint<X>` is not very good so open to suggestions on this, but from what I can see this is the data that's needed.

Alternatively:
```ts
type TileDataPoint<T> = {
  label: string
  data: T
};

type TileProps = {
  topText: string
  one: TileDataPoint<string[]>
  two: TileDataPoint<number | string[]>
};
```

Again though, `one` and `two` are poorly named.

eliascodes:
- `Polaroid` component to accept children
- `topText` prop to be split into two to support bolding part of it
- Additional prop to indicate different layout mode for `two.data`

astroash:
Updated interface suggestion:
```js
type TileDataPoint<T> = {
  label: string
  data: T
};

export type LayoutMode = 'number' | 'text';

type TileProps = {
  topText: string []
  left: TileDataPoint<string[]>
  right: TileDataPoint<number | string[]> & { layoutMode: LayoutMode }
  bottomText: string []
};
```

See #219 for working example

eliascodes:
`layoutMode` isn't used in #219

astroash:
> layoutMode isn't used in #219

😅 not yet, I was trying to throw something up quickly as I knew you were working on related work. This is a demo of what it would look like when set to `'number'`, `MiddleTopRightContainer` will return something different if set to `'text'`

eliascodes:
So I'm assuming the code will look something like:
```ts
if (props.layoutMode === 'number') {
  return formatNumberData(props.right.data);
} else if (props.layoutMode === 'text') {
  return formatTextData(props.right.data);
}
```

[...]

Issue #174, opened 2019-06-18
Switching to CSS grid


eliascodes:
The flexbox library we're using is annoying and limiting.

Investigate alternatives, specifically CSS Grid based alternatives (since it has a more convenient API for our typical use cases).

e.g.: https://styled-css-grid.js.org/

PR #219, opened 2019-07-04
📊 Dashboard data card component

astroash:
**Changes:**
- create DataCard element
- update Polaroid props
- import bold roboto

relates #192 #174


dashboard-app/src/features/dashboard/components/DataCard/util.ts
```diff
@@ -0,0 +1,20 @@
+
+export const sentenceCreator = (xs: string[]) => {
+  switch (xs.length) {
+    case 0:
+    case 1:
+      return xs;
+    case 2:
+      return [xs[0], 'and', xs[1]].map((x) => x.replace('and', ' and '));
```
eliascodes: isn't this the same as `[xs[0], ' and ', xs[1]]`, except with the benefit(?) that any strings in `xs` containing the substring `'and'` wont be affected?

dashboard-app/src/features/dashboard/components/DataCard/util.ts
```diff
@@ -0,0 +1,20 @@
...
+    case 3:
+    default:
```
eliascodes: This case is more complicated than it needs to be. This handles `xs.length > 1`:
```ts
xs.slice(0, -1).join(', ').concat(' and ').concat(xs.slice(-1).join())
```

dashboard-app/src/features/dashboard/Dashboard/index.tsx
```diff
@@ -16,6 +17,30 @@ const H1 = styled(Heading1)`
 /**
  * Component
  */
+
+const testNumberProps = {
```
eliascodes: I've pulled and had a look locally so this can be removed now 👍

astroash: Can we still rebase our PRs onto one another so the feature work is contained in one commit?

dashboard-app/src/features/dashboard/components/DataCard/util.ts
astroash: The above example returns a `string` rather than `string[]`. The issue I was having was inserting a comma element after every word. I can do a mash up of the two to remove the ternary

dashboard-app/src/features/dashboard/Dashboard/index.tsx
eliascodes: Yeah I've rebased onto yours, but haven't pushed yet

dashboard-app/src/features/dashboard/components/DataCard/util.ts
eliascodes: Maybe you can look at `intersperse` in ramda

eliascodes:
Would also be nice to have some component tests

dashboard-app/src/features/dashboard/components/DataCard/util.ts
eliascodes:
```ts
R.intersperse(', ', xs.slice(0, -1)).concat('and').concat(xs.slice(-1))
```

astroash:
> Would also be nice to have some component tests
This is a dumb component, not sure what would be tested.

eliascodes:
Closed in favour of #222

PR #222, opened 2019-07-10
📊 Data tiles feature

eliascodes:
FROM #219:
- create DataCard element
- update Polaroid props
- import bold roboto

relates #192 #174 

ALSO:
- Add utils fns for array/object processing
- Add fns to process API data into tile props
- Add error and loading boundary components!


dashboard-app/src/lib/ui/components/Boundaries/ErrorBoundary.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  componentDidCatch (error: Error, errorInfo: { componentStack: any }) {
+    if (process.env.NODE_ENV !== 'test') {
+      console.log(error, errorInfo);
+    }
+  }
+
+  renderError () {
+    return (
+      <div>
```
eliascodes: NOTE!

I haven't styled anything here because I don't know what we want this to look like.

dashboard-app/src/features/dashboard/components/DataCard/util.ts
```diff
@@ -0,0 +1,16 @@
+import { intersperse } from 'ramda';
```
eliascodes: This seems like its sufficiently generic to go into `twine-utils`?

dashboard-app/src/features/dashboard/Dashboard/index.tsx
```diff
@@ -25,46 +34,57 @@ const Dashboard: React.FunctionComponent<RouteComponentProps> = (props) => {
...
-                caption="See what projects volunteers spend their time doing."
-                callToAction="Coming soon"
-                picture="P"
-                disabled
-              />
-            </Col>
-          </Row>
-        </Col>
+        <LoadingBoundary isLoading={loading}>
+          <ThrowIfError error={error} />
```
eliascodes: This is kind of awkward. I've just imagined a solution that looks like this:

```ts
() => {
  return (
    <LoadingBoundary state={useDashboardStatistics()}>
		{
			(data) => (
				<Col>
					// ...
				</Col>
			)
		}
	</LoadingBoundary>
  );
}
```

In this example, `LoadingBoundary` takes this standard(ish) `{ loading, error, data }` object, and takes care of rendering loaders and error views if either `loading` or `error` are truthy. Once `data` is available, it's passed as a prop to the child component, which is rendered.

But tbh I don't know if we want to do this now.

dashboard-app/src/features/dashboard/components/DataCard/util.ts
astroash: Yeah, can do

dashboard-app/src/lib/ui/components/Boundaries/ErrorBoundary.tsx
```diff
@@ -0,0 +1,76 @@
...
+      isHttpError: isHttpError(error),
+      statusCode: propOr(0, 'code', error),
+      message: error.message,
+    };
+  }
+
+  componentDidCatch (error: Error, errorInfo: { componentStack: any }) {
+    if (process.env.NODE_ENV !== 'test') {
+      console.log(error, errorInfo);
+    }
```
astroash: Would be rad if we could sent the error to the server here. I'm leaning to think that these boundaries should be added as a separate PR so we can do styling n server changes in one.

dashboard-app/src/lib/ui/components/Boundaries/LoadingBoundary.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ColoursEnum } from '../../design_system';
+import { BeatLoader } from 'react-spinners';
+
+
+type LoadingBoundaryProps = {
+  isLoading?: boolean | (() => boolean)
+};
+
+
+class LoadingBoundary extends ErrorBoundary<LoadingBoundaryProps> {
```
astroash: Why does this extend `ErrorBoundary`? Does it need to use the `componentDidCatch` hook?

eliascodes: So that it acts as a boundary for errors as well as for loading states. If it doesn't extend `ErrorBoundary` I'd just have to re-write all that functionality again.

dashboard-app/src/features/dashboard/Dashboard/index.tsx
eliascodes: This is actually a bad idea because it breaks one of the rules of hooks: that they should be called at the top level. But I pushed something that does something similar.

dashboard-app/src/lib/ui/components/Boundaries/LoadingBoundary.tsx
astroash: Ah, I got confused with naming! I assumed `ThrowIfError` was the error boundary. Is there a reason that this components functionality is separate to`LoadingBoundary`?

eliascodes: Because an error boundary component cannot catch errors that it throws itself, the error must be thrown by a child component.

dashboard-app/src/lib/ui/components/Boundaries/ErrorBoundary.tsx
astroash: The text here isn't loading for me locally. I get a blank screen when running production build with the following console errors:
```
Error: ""
    er index.tsx:91
    React 8
        Zi
        Ro
        Ga
        Xa
        js
        Cs
        Ns
        Mr
react-dom.production.min.js:4408
    React 7
        ha
        callback
        oa
        ia
        Wa
        qa
        Ds
    unstable_runWithPriority scheduler.production.min.js:274
    React 5
        Ds
        js
        Cs
        Ns
        Mr
```

dashboard-app/src/lib/ui/components/Boundaries/LoadingBoundary.tsx
astroash: Is the aim to keep `ThrowIfError` indefinitely or hand over throwing responsibilities to internal function calls?

dashboard-app/src/lib/ui/components/Boundaries/ErrorBoundary.tsx
eliascodes: I get this:

Screenshot 2019-07-12 at 15 23 18

dashboard-app/src/lib/ui/components/Boundaries/LoadingBoundary.tsx
eliascodes: I think we need to keep it, but it should be internal to the boundary components and not used directly by any other components.

dashboard-app/src/lib/ui/components/Boundaries/ErrorBoundary.tsx
astroash: How are you creating an error?

eliascodes: Either by setting `error = new Error('oops')` in the component or by throwing from some component in the child component tree.

dashboard-app/src/features/dashboard/components/DataCard/util.ts
eliascodes: Are we doing this?

dashboard-app/src/lib/ui/components/Boundaries/ErrorBoundary.tsx
eliascodes: What should we do if there's an error? We could fall back to using the placeholders, but if there's an error in fetching the data on the dashboard there's probably gonna be one on the rest of the dashboard too, so not sure it makes sense to let them continue.

What I should probably do is create two error boundaries, one which renders something itself, and one which redirects (probs by rendering a `<Redirect>` component), and then we can do the same redirect behaviour that we currently have on API calls.

The former component can be used for failures in non-critical components where it's ok for us to just render some placeholder without the app becoming unusable.

dashboard-app/src/features/dashboard/components/DataCard/util.ts
astroash: yeah, feel free to make the changes

PR #223, opened 2019-07-11
🐛fix csv sorting

eliascodes:
~~**DO NOT MERGE**~~

Fix bad naming ~~before merging!~~ <- suggestions welcome

Related to #98

### DIFF
--- a/lib/twine-util/misc.ts
+++ b/lib/twine-util/misc.ts
@@ -1,0 +1,9 @@
+import { useWith, unapply, identity, equals, always, compose, cond, juxt, flip, T } from 'ramda'
+
+const equalsAndAlways = useWith(unapply(identity), [equals, always]);
+
+// toggle :: a -> b -> (* -> *)
+export const toggle = compose(
+  cond,
+  juxt([equalsAndAlways, flip(equalsAndAlways), always([T, identity])])
+) as <U> (x: U, y: U) => (z: U) => U;
--- a/dashboard-app/src/features/dashboard/ByActivity/index.tsx
+++ b/dashboard-app/src/features/dashboard/ByActivity/index.tsx
@@ -16,11 +16,10 @@ import { ColoursEnum } from '../../../lib/ui/design_system';
 import Errors from '../components/Errors';
 import useAggregateDataByActivity from './useAggregateDataByActivity';
 import { TabGroup } from '../components/Tabs';
-import { getTitleForDayPicker, toggleOrder } from '../util';
+import { getTitleForDayPicker } from '../util';
 import { useErrors } from '../../../lib/hooks/useErrors';
 import { TitlesCopy } from '../copy/titles';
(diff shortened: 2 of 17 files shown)


lib/twine-util/misc.ts
```diff
@@ -0,0 +1,9 @@
+import { useWith, unapply, identity, equals, always, compose, cond, juxt, flip, T } from 'ramda'
+
+const equalsAndAlways = useWith(unapply(identity), [equals, always]);
+
+// toggle :: a -> b -> (* -> *)
+export const toggle = compose(
```
eliascodes: Don't take this the wrong way, but this feels absurd. Why use 10(!) functions when one will do?
```ts
const toggle = <T>(left: T, right: T) => (x: T) => {
  if (x !== left && x !== right) throw new Error('wut');
  return x === left ? right : left;
}
```
or
```ts
const toggle = <T>(left: T, right: T) => (x: T) => {
  if (x !== left && x !== right) return left;
  return x === left ? right : left;
}
```

astroash: This was taken from ramda cookbook

https://github.com/ramda/ramda/wiki/Cookbook#toggle

eliascodes: I know, but that doesn't make it good lol

astroash: I like that it's composable. We can wrap one of your suggestions in `curry` if you want. Re the two options- It's worth having a discussion on how we want to handle errors in utility functions. I like functions throwing errors, however our other clienside apps are not designed to handle these cases currently.

eliascodes: > We can wrap one of your suggestions in curry if you want

That's fine. It just strikes me as bad engineering to generate more than 10 stack frames just to flip a value.

> It's worth having a discussion on how we want to handle errors in utility functions

I'm happy to default to throwing unless there's a good argument not to. You'd hope we'd be writing application code that can cope with a few thrown errors :)

astroash: > Which is to say, I think it should be the responsibility of application code to make sure everything doesn't fall over if something goes wrong in a util fn.

AFAIK this is not the case in either the visitor app or the volunteer app.

eliascodes: > AFAIK this is not the case in either the visitor app or the volunteer app

Yeah it's tech debt that we can should care of. The dashboard will be a little more robust b/c of Typescript.

astroash: Okay, so my concern is writing util functions in a manor that cannot be used across all of our apps. This involves us remembering which methods can and cannot be used where.

eliascodes: We're off topic from the original comment, let's move this discussion to a separate issue so we can close out this PR

astroash:
Force merging to avoid holding up QA

PR #237, opened 2019-07-29
🐛 fix dashboard tile truncation logic

### LINKED ISSUE #236: Truncated stats in dashboard tiles
Steps:  Using Twine Testing WITW org, View Time tile.  The data is/was setup so that there are 4 months with the same no of highest hours.
Expected: If more than three meet the criteria then an indication (…) is displayed that more months are included.  I would expect to see ""Most volunteer days were in January 2019, February 2019, May 2019...""
Actual: No ""..."" is displayed.

Steps:  Using Twine Testing WITW org, View Activities tile.  The data is/was setup so that there are 4 activities with the same no of highest hours.
Expected: If more than two activities have the same amount of time logged to them, display the top two alphabetically then an indication (…) is displayed that more activities are included.  I would expect to see ""Most popular activities were Office support, Other...""
Actual: No ""..."" is displayed.   Order is not top two alphametically.

Steps:  Using Twine Testing WITW org, View Volunteers tile.  The data is/was setup so that there are 4 volunteers with the same no of highest hours.
Expected: If more than three volunteers have the same amount of time logged (rounded to hours), display the first three alphabetically then an indication (…) that more volunteers are included in the list.  I would expect to see ""Top Volunteers Moly Mole, Mr Badger, Mr Toad...""
Actual: No ""..."" is displayed.   Order is not top three alphabetically.

eliascodes:
Fixes #236

Moved the truncation logic inside `DataCard` (mediated by props) because it handles the bolding internally. Alternatively we could make `DataCard` accept render props and move the logic outside, but this seems more complicated than necessary. The other alternative is to add more of the application logic inside the `DataCard` but this seems like a bad idea if we're trying to keep it as dumb as possible.

### Changes
- Add tests on `*StatsToProps` functions
- Modify `String.listify` to accept flag telling it not to insert `"and"` in the list -- defaults to the pre-existing behaviour
- Add another util to handle truncation
- `alternateBold` accepts an `offset` argument to avoid having to prepend an empty string to the argument sometimes
- Statistics functions now pass _all_ top values to the `DataCard` component, not just the first three, and the `DataCard` does the truncation.

### Testing Requirements
QA already passed

### Release Requirements
Can be released as soon as it is merged.

### Manual Deployment
- [ ] Dashboard App

### DIFF
--- a/dashboard-app/src/features/dashboard/Dashboard/useDashboardStatistics.ts
+++ b/dashboard-app/src/features/dashboard/Dashboard/useDashboardStatistics.ts
@@ -4,6 +4,7 @@ import { innerJoin, collectBy } from 'twine-util/arrays';
 import { useBatchRequest } from '../../../lib/hooks';
 import { CommunityBusinesses } from '../../../lib/api';
 import { findMostActive } from './util';
+import { NumberTileProps, TextTileProps } from '../components/DataCard/types';
 import Months from '../../../lib/util/months';
 
 
@@ -109,7 +110,7 @@ export default () => {
 
 const getCurrentMonth = () => moment().format(Months.format.abreviated);
 
-export const activityStatsToProps = (pts?: EqualDataPoints) => {
+export const activityStatsToProps = (pts?: EqualDataPoints): NumberTileProps => {
   if (! pts) {
     return {
       topText: ['During ', getCurrentMonth()],
@@ -137,7 +138,7 @@ export const activityStatsToProps = (pts?: EqualDataPoints) => {
       label: leftLabel,
       data: pts.labels,
       limit: 2,
-      placeholder: '...',
+      truncationString: '...',
     },
     right: {
       label: rightLabel,
@@ -147,7 +148,7 @@ export const activityStatsToProps = (pts?: EqualDataPoints) => {
 };
 
 
-export const volunteerStatsToProps = (pts?: EqualDataPoints) => {
+export const volunteerStatsToProps = (pts?: EqualDataPoints): TextTileProps => {
   if (! pts) {
     return {
       topText: ['During ', getCurrentMonth()],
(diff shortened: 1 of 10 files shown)


dashboard-app/src/features/dashboard/Dashboard/useDashboardStatistics.ts
```diff
@@ -63,7 +64,7 @@ export default () => {
     // most active months (12 months)
     setTimeStats(
       findMostActive(
-        collectBy((log) => moment(log.startedAt).format(Months.format.verbose), fullLogs)
+        collectBy((log) => moment(log.startedAt).format(Months.format.abreviated), fullLogs)
```
astroash: Why has this been changed?

dashboard-app/src/features/dashboard/Dashboard/useDashboardStatistics.ts
```diff
@@ -205,6 +210,8 @@ export const timeStatsToProps = (pts?: EqualDataPoints) => {
     left: {
       label: leftLabel,
       data: pts.labels,
+      limit: 3,
+      placeholder: '...',
```
astroash: Placeholder is a misleading name. This does not describe what this value is doing.

astroash:
@eliascodes can you update this PR in line with https://github.com/TwinePlatform/twine-monolith/issues/238#issuecomment-517657111

astroash:
@eliascodes can you update this PR in line with https://github.com/TwinePlatform/twine-monolith/issues/238#issuecomment-517657111

dashboard-app/src/features/dashboard/Dashboard/useDashboardStatistics.ts
eliascodes: This is the date format specified in the user stories (010A/B/C)

Issue #238, opened 2019-07-29
Release Pipeline rework


astroash:
Copied from slack:

> **[Current release pipeline]**
> - PR’s are reviewed by devs and merged into staging environment
> - Staging environment is reviewed by QA & any changes requested are merged into staging
> - On approval of PO changes are deployed to production environment
>
>**[Motivation]**
> The above works if we are working one feature at a time. However if various pieces of work take places simultaneously we end up in a situation where unrelated changes are on staging environment and *we are unable to have granular control over what is deployed to production*.
>
> **[Questions]**
> Me and Elias can propose a technical solution for release pipelines but need to confirm the following questions:
> - Do we want to acceptance test minor changes?
>   - Will this run through QA/PO approval or can dev release when it has gone through code review
> - To what level do we want to run regression testing?
>   - Is this being done on new changes?
>   - Do we want to define critical user paths so we have a light weight & in depth test packs for regression testing
>   - Do we want a monthly/bi-monthly regression test of all apps?
> - How do we want to test not user facing changes ie file restructures/dependency updates/refactors

New setup to be discussed

astroash:
Suggested solution:

- Return to production being manually deployed from github master
- Review apps created for each PR (manually/mirrored apps/different buildpack https://github.com/TwinePlatform/twine-monolith/issues/22#issuecomment-514274614)
- Feature branches being created for scheduled release work
- Hotfix branches being created and merged directly into master

eliascodes:
> Return to production being manually deployed from github master

Does this mean we have a staging/dev branch again, like in the diagram I showed you, or are you imagining we wouldn't need this if we have feature branches?

> Review apps created for each PR

I'm happy to go ahead with this but I can imagine that if there's no automated way to do this it will become quite burdensome.

> different buildpack

Specifically on this, I've had a look and it doesn't seem to do anything special. Our current buildpack should support yarn workspaces already.

Everything else sounds fine.

astroash:
> Does this mean we have a staging/dev branch again
I cannot see any use for them in our setup. Can you think of anything I am overlooking?

> if there's no automated way to do this it will become quite burdensome.
Hosted review apps will only be needed if other team members are reviewing work. I imagine this will be less often with the impending team structure changes. 

> Our current buildpack should support yarn workspaces already.
We cannot have review apps as we need `app.json` to live in the root directory. I've not looked into the buildpack, but I wonder whether it has another way to support this as it's made by heroku.

[...]

PR #247, opened 2019-08-01
📜 reporting: grab users instead of only volunteers

eliascodes:
Grabbing only volunteers in the raw dump means "creator" information in cases where CB_ADMIN users are creating logs on behalf of VOLUNTEER users is left blank.

The data dump in GDrive has already been updated with this change.

**Testing Requirements:** None

### Manual Deployment
- [ ] **API**


astroash:
@eliascodes can you update this PR in line with https://github.com/TwinePlatform/twine-monolith/issues/238#issuecomment-517657111

PR #256, opened 2019-08-05
🕸 add webhook endpoint for heroku releases

astroash:
relates #238 

### Changes
- add heroku strategy for authentication
- add webhook endpoint sending devs emails
- add devs emails to env var
- add template id for postmark
- NOTE: not tested as this is a dev facing action. Can test if wanted

### Testing Requirements
- not user facing test in production

### Release Requirements
- no release restrictions, deploy when ready

### Manual Deployment
- [ ] api

### Additional steps

- [ ] add webhooks to all staging apps
- [ ] add new envs to api staging on heroku

### DIFF
--- a/api/config/config.schema.ts
+++ b/api/config/config.schema.ts
@@ -82,7 +82,7 @@ export default Joi.object({
   email: Joi.object({
     postmarkKey: Joi.string().required(),
     fromAddress: Joi.string().email().required(),
-    developers: Joi.array().items(Joi.string()).length(2),
+    developers: Joi.array().items(Joi.string().email()),
   }).required(),
   auth: Joi.object({
     schema: {
--- a/api/config/config.schema.ts
+++ b/api/config/config.schema.ts
@@ -82,6 +82,7 @@ export default Joi.object({
   email: Joi.object({
     postmarkKey: Joi.string().required(),
     fromAddress: Joi.string().email().required(),
+    developers: Joi.array().items(Joi.string()).length(2),
   }).required(),
   auth: Joi.object({
     schema: {
@@ -107,5 +108,10 @@ export default Joi.object({
   }).required(),
   qrcode: Joi.object({
     secret: Joi.string().min(32).required(),
+  }).required(),
+  webhooks: Joi.object({
+    heroku: Joi.object({
+      secret: Joi.string().min(32).required(),
+    }).required(),
   }).required(),
 });
--- a/api/src/services/email/index.ts
+++ b/api/src/services/email/index.ts
@@ -1,5 +1,5 @@
 import EmailDispatcher from './dispatcher';
-import { EmailTemplate, Templates } from './templates';
+import { EmailTemplate, Templates, WebhookEmailTemplates } from './templates';
 import { Config } from '../../../config';
 import { User, CommunityBusiness } from '../../models';
 import { RoleEnum } from '../../models/types';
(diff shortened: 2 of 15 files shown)


api/config/config.schema.ts
```diff
@@ -73,6 +73,7 @@ export default Joi.object({
   email: Joi.object({
     postmarkKey: Joi.string().required(),
     fromAddress: Joi.string().email().required(),
+    developers: Joi.array().items(Joi.string()).length(2),
```
eliascodes:
```suggestion
    developers: Joi.array().items(Joi.email()).length(2),
```

api/src/webhooks/v1/heroku/onStagingRelease.ts
```diff
@@ -0,0 +1,38 @@
...
+      description: 'Sends email when heroku app releases to staging',
+      auth: {
+        strategy: 'herokuWebhook',
+      },
+      validate:
+      {
+        payload: {
+          data: {
+            app: Joi.object({
+              name: Joi.string(),
```
eliascodes: If we're sure these parameters are always sent, lets make the `.required()`?

api/src/webhooks/v1/heroku/onStagingRelease.ts
```diff
@@ -0,0 +1,38 @@
+import * as Hapi from '@hapi/hapi';
+import * as Joi from '@hapi/joi';
+
+
+const routes = [
+
+  {
+    method: 'POST',
+    path: '/heroku/on-staging-release',
```
eliascodes: If this is in a `v1` folder (i.e. we're versioning the webhooks), we should probably add some versioning to the URL too.

eliascodes:
Also this is the email I received when you sent out a test, looks like the app name might not be being passed down? Not sure if this is fixed now:

Screenshot 2019-08-06 at 19 13 22

api/src/webhooks/v1/heroku/onStagingRelease.ts
astroash: That's done when the routes are registered, same as the api. Routes are all prefixed with `/webhooks/v1`

api/config/config.schema.ts
astroash: If using this, what do you recommend for the defaults? Currently the defaults are not emails, so if the env is not set, postmark will silently error

astroash:
> Also this is the email I received when you sent out a test, looks like the app name might not be being passed down?

Yeah I tried passing the values in the template but it didn't. Seemed not worth spending time on since it was just for us.

api/config/config.schema.ts
eliascodes: You can leave the defaults as `[]`, then if you `map` over the array to create the inputs for `sendBatch` instead of directly indexing it'll just pass an empty array to postmark (which should hopefully do nothing?)

api/src/services/email/index.ts
```diff
@@ -24,8 +24,36 @@ export type EmailService = {
...
+        to: cfg.email.developers[0],
+        templateId: WebhookEmailTemplates.ON_HEROKU_RELEASE,
+        templateModel: {
+          app_name: appName,
+          product_name: 'Twine Platform',
+        },
+      },
+      {
+        from: cfg.email.fromAddress,
+        to: cfg.email.developers[1],
```
eliascodes: Is there any reason we only ever want to send two emails here? If not, suggest we just `map` over the emails?

Issue #246, opened 2019-08-01
Corrupted logs cannot be synced


eliascodes:
Discovered as a result of one of our users being unable to sync their data. Discovered this was because one of the logs that had been saved offline was corrupted; specifically, the `startedAt` date was an invalid date string in the format `"undefined HH:mm:ss"`. User was unable to manually fix this date string because the logs were also marked as deleted, and therefore don't appear in the app. 

1. Something is going wrong with the date string generation somewhere
2. Might be a good idea for users to be able to see deleted logs somehow
3. Users should be able to manage what's kept in offline storage (and clear it if necessary)
4. The way offline storage currently works is byzantine and difficult to maintain (see TwinePlatform/twine-app#90)

eliascodes:
This is being resolved by ignoring corrupt logs. In the update of the native app (#327) we'll fix this by having a more robust offline syncing system.

PR #245, opened 2019-07-31
[WIP] 🚑 HOTFIX allow invalid start dates through validation 

eliascodes:
**DO NOT MERGE**

...but then ignore them from rest of handler.

related to #246

PR #227, opened 2019-07-16
✨ Server-side sessions

eliascodes:
fixes #207 

~80% of the diff is renaming and config changes.

Manually tested with visitor, volunteer and dashboard apps.

### Changes
- Use `@hapi/yar` and `@hapi/catbox-redis` to setup server-side session cache
- Create `session_cookie` authentication schema to validate requests using the session cache
- Re-write the `standard` authentication strategy to use the `session_cookie` schema and rely on `yar`
- Add config for the new auth system and the session cache
- Add a bit of documentation
- Update tests
- Remove pointless `src/routes` file
- Remove `hapi-auth-jwt2` and `jsonwebtoken`

### Testing Requirements

- [x] **Visitor app**

- Sign in
- Register visit
- Go to admin area
- Visit an admin page
- Sign out
- Try to (unsuccessfully) navigate directly to admin area (by typing in URL)

- [x] **Volunteer app**
- Sign in
- Create log
- Update user info
- Sign out

- [x] **Dashboard**
- Sign in
- Click one of the tiles
- Check data loads
- Sign out

### Release Requirements
- Thursday 8th August

### Manual Deployment
- [ ] **API**
- [ ] **Visitor app**
- [ ] **Dashboard**

PR #259, opened 2019-08-08
🚑 ignore invalid date-strings from synced logs

eliascodes:
related to #246 

### Changes
- Mainly an attempt to prevent this endpoint from returning non-200 status codes where reasonable to avoid users of the volunteer app being trapped in offline mode ☹️
- re-implement the temporary fix in #245 
  - allow logs with invalid timestamps for `startedAt` and `deletedAt` through validation but ignore them when actually syncing
- Remove sync operation from transaction: if only some logs can be synced, they should be, others ignored
- Endpoint responds with `{ synced: Number, ignored: Number }` to avoid transaction problems in tests
- Ignored logs are stored as JSON blobs in temporary table `invalid_synced_logs_monitoring`.
- Add some `Scopes` utility functions (with a little added validation)
- Add a `Promises.some` utility function: like `Promise.all` but resolves no matter what and stores the error objects of each rejected promise.

### Testing Requirements
- [x] Volunteer app
  - With changes:
    - Use web-interface
    - Using the dev tools, disable network
    - Add more than one log (app should go into offline mode)
    - Using dev tools, manually edit the offline cache (local storage) to corrupt one of the logs' timestamps
    - Re-enable network, then try to sync logs
    - Expect app to leave offline mode and successfully sync stored logs, except for corrupted log, which should be ignored.
  - Without changes:
    - Steps as above, but app should be unable to leave offline mode

### Release Requirements
Release informally agreed at planning meeting (5/8/19)

### Manual Deployment
- API

### DIFF
--- a/api/src/api/v1/community_businesses/volunteer_logs.ts
+++ b/api/src/api/v1/community_businesses/volunteer_logs.ts
@@ -14,6 +14,7 @@ import {
   volunteerLogDuration,
   volunteerProject
 } from './schema';
+import * as Scopes from '../../../auth/scopes';
 import Roles from '../../../models/role';
 import { VolunteerLogs, Volunteers, VolunteerLog } from '../../../models';
 import { getCommunityBusiness } from '../prerequisites';
@@ -29,6 +30,8 @@ import { requestQueryToModelQuery } from '../utils';
 import { query } from '../users/schema';
 import { Credentials as StandardCredentials } from '../../../auth/strategies/standard';
 import { RoleEnum } from '../../../models/types';
+import { PermissionLevelEnum } from '../../../auth';
+import { AccessEnum, ResourceEnum } from '../../../auth/types';
 
 
 const ignoreInvalidLogs = (logs: SyncMyVolunteerLogsRequest['payload']) => {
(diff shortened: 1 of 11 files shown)


astroash:
tested and received a `403` error

eliascodes:
the 403 is because it’s a CB_ADMIN user. The sync endpoint needs “own” or “sibling” scopes, CB_ADMIN only has “child”. Testing with a VOLUNTEER or VOLUNTEER_ADMIN should work.

astroash:
> Testing with a VOLUNTEER or VOLUNTEER_ADMIN should work.

The volunteer app was designed to work with `CB_ADMIN`s. The scoping on this route needs to be changed.

eliascodes:
looks like I need to rebase to avoid capturing changes from #227 in here

astroash:
@eliascodes I'm still getting a 403
Screen Shot 2019-08-14 at 09 26 39

eliascodes:
Not sure what account you're testing with, I just tested it with `demo@` on my local clone (which is a `CB_ADMIN`) and it worked.

astroash:
@eliascodes can you outline what changes are new. I had previously approved this or to merge when you were ready.

eliascodes:
All the bullet points in the PR description except for the second one describe stuff that's been added since your last review.

eliascodes:
@astroash did you want to dismiss your old review?

api/src/api/v1/community_businesses/volunteer_logs.ts
```diff
@@ -398,51 +400,52 @@ const routes: Hapi.ServerRoute[] = [
       }
 
       try {
-        await knex.transaction(async (trx) =>
-          Promise.all(payload.map(async (log) => {
-            const existsInDB = log.hasOwnProperty('id');
-            const shouldUpdate = existsInDB && !log.deletedAt;
-            const shouldDelete = existsInDB && log.deletedAt;
+        const results = await Promises.some(payload.map(async (log) => {
```
astroash: I think it would be more readable if this was all pulled out to a model method. There's a lot going on in this handler

api/src/api/v1/community_businesses/volunteer_logs.ts
```diff
@@ -362,11 +367,11 @@ const routes: Hapi.ServerRoute[] = [
 
       // Ignore invalid logs silently because of
       // https://github.com/TwinePlatform/twine-monolith/issues/246
-      const payload = ignoreInvalidLogs(_payload);
+      const { valid: payload, invalid } = ignoreInvalidLogs(_payload);
```
astroash: why is this method only used in the `sync` handler, couldn't this error happen on POST/PUT methods too?

eliascodes: I'm not sure because I haven't spent any time looking into why this timestamp corruption might have happened. In principle, if the timestamp is corrupted when adding/editing, the app would fall into offline mode, and the user would have to sync.

api/src/api/v1/community_businesses/volunteer_logs.ts
eliascodes: I agree with this, but I'd like to merge now, and then work on this immediately afterwards (just so we're not delaying deployment).

astroash: @eliascodes can you create an issue to outline the changes proposed n I'll merge this now.

PR #261, opened 2019-08-08
Test yarn 2

astroash:
fixes ...

### Changes
-
-

### Testing Requirements
- [ ] 
  - 
  - 

### Release Requirements
- 

### Manual Deployment
- [ ]

PR #269, opened 2019-08-13
👀 implement monitoring of user sessions

### LINKED ISSUE #255: Implement usage monitoring
https://docs.google.com/document/d/1vFvb6pLorSCOiLNSG-3mF2FXJ53T_atYNHOBi4AdtWY/edit

eliascodes:
fixes #255 

### Changes
- Add step to release phase to configure redis
- Add `user_session_record` table and `UserSessionRecords` collection
- Update `user_session_record` when users:
  - Login
  - Make authenticated requests to the API
  - Logout
  - Session expire
- Add `ioredis` dependency
- Add listener when redis session keys expire (and update `user_session_record`)
- Add session monitoring assertions to `auth_flow` test
- Add `Promises.silent` util which ignores rejections for auxiliary actions
- Remove calls to insert into `user_account_active_day` and `login_event` -- this is can be calculated from `user_session_record`

### Testing Requirements
- [x] Dashboard
  - Login (then check `user_session_record` is created)
  - Navigate to various pages (then check `user_session_record.referrers` is updated)
  - Wait for the session to expire (then check `user_session_record.session_end_type = 'expired'`)
  - Do the above flow again but log out instead
- [x] Visitor app
  - As for dashboard
- [x] Volunteer app
  - As for dashboard

### Release Requirements
To be confirmed

### Manual Deployment
- API


api/database/redis/configure.ts
```diff
@@ -0,0 +1,20 @@
...
+  const config = _config ? _config : getConfig(process.env.NODE_ENV);
+  const client = new Redis(config.cache.session.options.url);
+
+  // Enable keyspace events (see docs/sessions#redis)
+  await client.config('set', 'notify-keyspace-events', 'AKE');
+  await client.quit();
+};
+
+
+if (require.main === module) {
```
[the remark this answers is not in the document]
eliascodes: If the file is invoked directly (e.g. `node file.js` or `ts-node file.ts`), this condition is true and the branch runs.

Basically I want to be able to run the file directly in the heroku release phase, and also require it in the jest `global_setup.ts` file.

api/database/migrations/sql/20190808_create_user_session_monitoring_table.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE user_session_record (
+  user_session_record_id SERIAL NOT NULL UNIQUE,
+  user_account_id        INT NOT NULL,
+  organisation_id        INT NOT NULL,
+  session_id             VARCHAR NOT NULL UNIQUE,
+  session_end_type       VARCHAR,
+  referrers              JSONB NOT NULL,
+  started_at             TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  ended_at               TIMESTAMP WITH TIME ZONE,
+  created_at             TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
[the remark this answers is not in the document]
astroash: @eliascodes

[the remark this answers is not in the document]
eliascodes: I removed it in my latest changes, now it's just `started_at`.

lib/twine-util/promises.ts
```diff
@@ -33,3 +33,7 @@ export const find = async <T>(fn: (a: T) => Promise<boolean>, xs: T[]): Promise<
 export const some = async <T>(ps: Promise<T>[]) => {
   return Promise.all(ps.map((p) => p.catch((error) => error)));
 };
+
+export const silent = <T>(p: Promise<T>): void => {
```
astroash: Peer package installing is still broken. I could only get this working by traversing into this repo, building locally, then trashing & reinstalling `api`'s dependencies

eliascodes:
> This is not working for me

This is probs because I left the session expiry for the dev environment to 10 seconds (I was manually testing expiration monitoring). I've updated it to 1 day. But if that's not it, make sure you've also run migrations and got a redis server instance running.

lib/twine-util/promises.ts
eliascodes: I added the postinstall script back (not sure when that was removed), but might be worth resurrecting #261

api/database/migrations/sql/20190808_create_user_session_monitoring_table.sql
```diff
@@ -0,0 +1,26 @@
...
+ * Create user session monitoring table
+ */
+
+CREATE TABLE user_session_record (
+  user_session_record_id SERIAL NOT NULL UNIQUE,
+  user_account_id        INT NOT NULL,
+  organisation_id        INT NOT NULL,
+  session_id             VARCHAR NOT NULL UNIQUE,
+  session_end_type       VARCHAR,
+  referrers              JSONB NOT NULL,
```
astroash: This is a list of clientside routes that the session has accessed? It may be worth renaming this field to clearly describe this rather than being tied to request headers.

Also it only tells us about single access rather than quantity, not sure this data is useful in this format.

eliascodes: > Also it only tells us about single access rather than quantity, not sure this data is useful in this format.

It's useful but it can answer fewer questions than storing (for example) the number of times each page was accessed. More complex quantities/statistics were not specified in the user story, but tbh I can see it being potentially better for us to just store all the request headers for each request (stripping out useless stuff) and then worrying about what can be calculated from that later.

If that's the way forward I'll rename this column to `request_headers`, and it'll just be a `Dictionary<{}>[]`.

eliascodes:
> logging into two apps with the same account will create to separate entries

This sounds like it would only be possible if you're using two different browsers. If that's the case, then it's desired behaviour, since we're tracking sessions, not accounts. Tying sessions together is the work of data analysis.

eliascodes:
> logging into two apps with the same account will create to separate entries

So ultimately I think this is fine. If we decide to handle both those "sessions" together, we can do so in data analysis, but at least we'll have the choice, whereas we won't if we combine them now.

eliascodes:
So there's a problem here that I've discovered during deployment to staging: the heroku-redis add-on [does not allow keyspace notifications to be enabled](https://help.heroku.com/5BP0E8RC/how-to-enable-notify-keyspace-events-on-redis).

This means that the listener we attach to respond to session expiry events won't be executed, and the `user_session_record` rows won't be updated with `'expired'` entries (the `session_end_type` will just be left blank).

To be clear, this has no impact on session management itself, only on monitoring.

On reflection, this isn't really a big deal. Those rows would have remained blank until session expiry anyway, and we only need those records when performing data analysis. So we have a couple options:
1. When performing data exports of the session records, auto-fill `session_end_type` as `expired` for any row whose `created_at` is further back than the session TTL.
2. Run a cron job to fill in `expired` on a regular (daily/weekly?) basis.
3. Find a different redis provider that allows keyspace events.

I'm going with option (1) right now since it's zero additional work. We can do (3) at any stage. If we choose to never do (3) we can get rid of `standard/monitoring.ts`.

So I'm going to merge this, and we can address these options later.

PR #304, opened 2019-09-03
🚸 disable signup button when request is pending

astroash:
fixes https://github.com/TwinePlatform/twine-visitor/issues/553

### Changes
- disable button and add disabled styling
- refactor all references of `status`

### Testing Requirements
- [ ] **Visitor App**
  - sign up as new user who already exists
  - observe button being disabled 

### Release Requirements
- release when ready

### Manual Deployment
- Visitor App


visitor-app/src/shared/components/form/base.js
```diff
@@ -102,6 +102,17 @@ export const SecondaryButton = styled(Button)`
   border-radius: 0;
 `;
 
+export const DisabledButton = styled(Button)`
```
eliascodes: Sounds like this button is always going to be disabled, so we should probably use `styled(Button).attrs((props) => ({ disabled: true }))`

But also, do we want a separate component for a button that is disabled, or do we want buttons that can be disabled? e.g. using the `disabled` attribute to change the styling?

astroash: I had a play with doing this mainly to transition the background, but then discovered that `background: linear-gradient(...` isn't transitionable.

Can move back to this setup if you want.

eliascodes: What if instead of transitioning `linear-gradient` you transition `opacity`?

astroash: Yeah you can do that, then update the z-index on a delay:
https://css-tricks.com/transitioning-gradients/

Seemed a bit over the top for a small feature.

eliascodes: Oh I didn't mean using opacity to do a linear-gradient transition, I just meant that the visual indication that the button is disabled would be it going slightly transparent. But if it looks crappy then don't worry about it 🤷🏼‍♂️

astroash: Yeah, that would not look good. Grey is standard disabled.

eliascodes: OK cool. Well given that we don't have a transition right now anyway (since we're rendering a different component), I think let's go back to having button components that can be disabled, it's cleaner.

visitor-app/src/shared/components/form/base.js
```diff
@@ -74,7 +74,8 @@ export const Button = styled.button`
   letter-spacing: 0.2em;
 `;
 
-export const PrimaryButton = styled(Button)`
+const standardPrimaryButtonStyling = `
+  cursor: pointer;
```
eliascodes: Let's promote this into `Button`? Seems like it should be standard across all buttons.

visitor-app/src/shared/components/form/base.js
```diff
@@ -91,6 +92,24 @@ export const PrimaryButton = styled(Button)`
...
 
+const disabledButtonStyling = `
+  color: ${colors.white};
+  background-color: ${colors.dark}; /* Fallback */
+  background: linear-gradient(
+    0,
+    ${rgba(colors.dark, 0.75)} 0%,
+    ${colors.dark} 100%
+  );
+  border-radius: 0;
```
eliascodes: Is this deliberate? Why is border radius changing on disable?

visitor-app/src/shared/components/form/base.js
```diff
@@ -91,6 +92,24 @@ export const PrimaryButton = styled(Button)`
...
+`;
+
+export const PrimaryButton = styled(Button)`
+  ${props => props.disabled
+    ? disabledButtonStyling
+    : standardPrimaryButtonStyling
+}
```
eliascodes: indentation

astroash: Linter config

visitor-app/src/shared/components/form/base.js
```diff
@@ -72,9 +73,10 @@ export const Button = styled.button`
   font-size: ${fonts.size.base};
   color: ${colors.black};
   letter-spacing: 0.2em;
+  border-radius: 0;
```
eliascodes: `border-radius` should not be 0 on these buttons.

visitor-app/src/visitors/pages/SignUp/signup_form.js
```diff
@@ -122,7 +123,7 @@ const signupForm = (props) => {
 
               <ErrorText show={props.errors.ageCheck}>{props.errors.ageCheck}</ErrorText>
             </div>
-            <SubmitButton type="submit">CONTINUE</SubmitButton>
+            <SubmitButton type="submit" disabled={props.status === status.PENDING}>CONTINUE</SubmitButton>
```
eliascodes: Maybe we should also change the button text to "Please wait" or something like that when the request is pending?

astroash: I think that will look jarring if there is a short request time and a validation error.

I can add a spinner internal to the button if you want. I started playing with one before. I was thinking a bounce loader would work:
https://www.react-spinners.com/

eliascodes: Let's see what it looks like I guess. If it looks ok then let's go for it. Why did you think bounce loading instead of beat loader? The latter would make it seem more consistent with the other loaders?

visitor-app/src/shared/components/form/base.js
astroash: Previously it was set on `SecondaryButton` and `PrimaryButtonNoFill`. Is this correct? I could not see a pattern for where it was and wasn't set so assumed it was meant to be on all of them.

eliascodes: Yeah basically the submit/form buttons have non-zero border-radius but the ones that are used on the admin dashboard and the landing page have zero border radius.

visitor-app/src/visitors/pages/SignUp/signup_form.js
astroash: Options:

button_1

button_2

button_3

button_4

eliascodes: IMO the grey ones look better, leaning towards the first one (only one dot), but no strong opinion

astroash: Opted for single loader in yellow as it complements the switch better:

button_5

visitor-app/src/visitors/pages/SignUp/SignUpForm/sideCopy.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
```
eliascodes: If this is a component the filename should be `PascalCase`

astroash: We're not following this convention in this repo. Happy to move forward like this though.

Issue #32, opened 2019-04-25
tslint deprecated


eliascodes:
https://medium.com/palantir/tslint-in-2019-1a144c2317a9

PR #307, opened 2019-09-03
🚸 display overlay when no data in range

eliascodes:
Found this stuff in the course of working on #32 but thought it should be split out.

### Changes
- Display overlay when `data` is empty
- Rename `noActiveLegendText` to `overlayText` since it is also triggered when there's no data in the selected range.
- Fix bug found in `logsToAggregatedData` where an empty range sometimes causes a crash

### Testing Requirements
- [x] Dashboard
  - Go to `/time` or `/volunteer`
  - Select a date range with no data
  - Observe overlay message

### Release Requirements
- Minor change, release when ready

### Manual Deployment
- Dashboard

### DIFF
--- a/dashboard-app/src/features/dashboard/components/StackedBarChart/index.tsx
+++ b/dashboard-app/src/features/dashboard/components/StackedBarChart/index.tsx
@@ -31,10 +31,31 @@ interface Props {
   defaultSelection: boolean;
 }
 
+
+/*
+ * Helpers
+ */
+
+const getOverlayText = (data: AggregatedData, legendItemsActive: boolean): [boolean, string] => {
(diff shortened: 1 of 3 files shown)


dashboard-app/src/features/dashboard/components/StackedBarChart/index.tsx
```diff
@@ -106,3 +103,19 @@ const StackedBarChart: FunctionComponent<Props> = (props) => {
 
 export default StackedBarChart;
 
+
+const getOverlayText = (data: AggregatedData, legendItemsActive: boolean): [boolean, string] => {
```
astroash: This relies on hoisting. Can this be moved to the top or to another folder.

PR #298, opened 2019-08-27
♻️ [FEATURE] Volunteer app update

eliascodes:
**DO NOT SQUASH**
**REBASE WHEN MERGING**

### Changes
- See commits

Post-review:
- Move "you haven't added any logs" message 
- Fix datepicker limits for admins
- Change "error" popup to "alert"
- Add postcode field to profile page
- Hide loading spinner when error occurs on editing project name

### Testing Requirements
- [x] Volunteer app
  - Build and release internal testing versions of app for Android/iOS
  - Manual testing recapping testing steps in #288, #292, #293 

#### 1. From #288 
- a. On Login, observe login page has placeholders removed and input boxes enlarged
- b. Login as VOLUNTEER
- c. Observe dashboard has copy changed as per recommendations in #278
- d. Observe "Projects" input moved above "Activities" input on "New Log" and "Edit Log" views
- e. Log out
- f. Observe Twine logo has been removed from register page
- g. Log in as VOLUNTEER_ADMIN
- h. Observe nav bar has the the copy change as per recommendations in #278

#### 2. From #292 
- a. Try to sync logs successfully after having been in offline mode (e.g. by trying to add a log while in airplane mode)

#### 3. From #293 
- a. Register as a new volunteer with an invalid post code (error message)
- b. Register as a new volunteer with a valid post code (success)
- c. Try to enable location reminders in app and check that an alert is displayed telling user to enable location services. Doesn't disappear until "OK" button is pressed

#### 4. From TwinePlatform/twine-app#33
- a. Log in as VOLUNTEER_ADMIN
- b. Attempt to edit a log that was created before the beginning of the current month

### Release Requirements
- PO Approval Required

### Manual Deployment
- Volunteer app
  - Android Play Store
  - iOS App Store
- API

PR #314, opened 2019-09-04
🐛 fix processing when trying to sync multiple logs stored offline

### LINKED ISSUE #313: Error when trying to sync multiple logs
Found in the latest build:
- Attempt to create a log for multiple volunteers at the same time
- With network disabled, press "save"
- Once in offline mode, re-enable network
- Attempt to sync logs
- Sync fails

This is because in the failure case on the new-logs-controller, we store the form data in local storage instead of the processed payload.

eliascodes:
fixes #313

### Changes
- Store processed payload data in offline cache instead of unprocessed form data

### Testing Requirements
- [x] Volunteer app
  - With network disabled, attempt to create a log for multiple users at the same time
  - When saving fails, enable network and attempt to sync
  - Syncing should succeed

### Release Requirements
- Bundle with #298 

### Manual Deployment
- Volunteer app

### DIFF
--- a/volunteer-app/www/js/controllers/new-log-controller.js
+++ b/volunteer-app/www/js/controllers/new-log-controller.js
@@ -397,7 +397,15 @@
 							$ionicLoading.hide();
 
 							// push to offline data and mark as 'needs_pushing'
-							$scope.newLogOffline($scope.formData, true, 'Log saved offline.');
+							if (payloads) {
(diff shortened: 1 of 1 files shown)


astroash:
@eliascodes has #298 been deployed with these changes?

eliascodes:
Yes, I tested this change myself on the internal testing version.

Issue #302, opened 2019-09-01
Heroku Redis doesn't support keyspace events


eliascodes:
Discovered in #269 

[Comment](https://github.com/TwinePlatform/twine-monolith/pull/269#issuecomment-526939383) reproduced here for convenience:

> So there's a problem here that I've discovered during deployment to staging: the heroku-redis add-on [does not allow keyspace notifications to be enabled](https://help.heroku.com/5BP0E8RC/how-to-enable-notify-keyspace-events-on-redis).

> This means that the listener we attach to respond to session expiry events won't be executed, and the `user_session_record` rows won't be updated with `'expired'` entries (the `session_end_type` will just be left blank).

> To be clear, this has no impact on session management itself, only on monitoring.

> On reflection, this isn't really a big deal. Those rows would have remained blank until session expiry anyway, and we only need those records when performing data analysis. So we have a couple options:
> 1. When performing data exports of the session records, auto-fill `session_end_type` as `expired` for any row whose `created_at` is further back than the session TTL.
> 2. Run a cron job to fill in `expired` on a regular (daily/weekly?) basis.
> 3. Find a different redis provider that allows keyspace events.

> I'm going with option (1) right now since it's zero additional work. We can do (3) at any stage. If we choose to never do (3) we can get rid of `standard/monitoring.ts`.

PR #316, opened 2019-09-09
🦸🏻‍♀️ Minor auth-related renaming

eliascodes:
As discussed during our walkthrough together

### Changes
- Add comments on important config variable
- Name variables instead of using `...args`
- Rename `Credentials.get` to `Credentials.create`
- Remove session expiry monitoring entirely (see #302)

### Testing Requirements
N/A

### Release Requirements
N/A

### Manual Deployment
N/A


api/src/api/v1/auth.ts
```diff
@@ -38,8 +38,6 @@ export default async (server: Hapi.Server) => {
    * Standard strategy
    */
   server.auth.strategy('standard', 'session_cookie', { validate: standardStrategy.validate });
-  const cleanup = standardStrategy.monitorSessionExpiry(knex, cache.session.options.url);
```
astroash: Why has this been removed?

eliascodes: Remember we spoke about removing this because of #302 and the fact that it was potentially the cause of flaky tests on CI?

Issue #211, opened 2019-07-02
🐏 ramda typing woes


astroash:
<img width="869" alt="Screen Shot 2019-07-02 at 15 35 08" src="https://user-images.githubusercontent.com/14337958/60509863-03fc0980-9cdf-11e9-99d1-efd88c8cb1b5.png">

eliascodes:
This is fixed in #209 and #210

eliascodes:
Happening again on #227

astroash:
I genuinely think that yarn workspaces will solve this issue as it hoists packages to the root level

eliascodes:
Yeah this has been annoying enough now, I'm happy to try moving to yarn.

PR #333, opened 2019-09-23
😱 fix travis & coverage config

### LINKED ISSUE #331: 👷🏻‍♀️ speed up CI with travis matrix
Probs the thing we need to run all our repos tests in parallel to speed up CI

https://docs.travis-ci.com/user/build-matrix/

### LINKED ISSUE #344: 👷‍♂️ travis playing up
CI is not acting as we intend:
- not failing on linting issues
- not failing on build issues

This is potentially an issue with our travis script implementation

eliascodes:
fixes #344 
fixes #331 
related to #211 

Uh-oh: 
- the filename processing in the bash-script was wrong, so we were only reporting twine-util coverage statistics
- also the loop in the `.travis.yml` doesn't exit with non-zero exit code when tests fail, so all builds since that change was made have potential errors

### Changes
- Fix filename processing in `cc-report.sh`
- Use build matrix to parallelise tests across apps
- Fix some tests/typings that were left over from old PRs not being checked by travis properly
- upload coverage reports to S3 before merging them together
- (potentially) permanently fix build issues with ramda type conflicts

### Testing Requirements
- [x] Travis
  - Look at the build log -- does it actually run and pass tests?
  - Are there any failures which don't cause the build to fail?
- [x] Codeclimate
  - Look at the coverage report -- does it actually make sense

### Release Requirements
N/A

### Manual Deployment
N/A


.travis.yml
```diff
@@ -17,25 +26,46 @@ services:
...
+      services: skip
+      install:
+        - pip install --user awscli
+      before_script:
+        - bash $TRAVIS_BUILD_DIR/.travis/cc_download.sh
+      script:
+        - aws s3 cp s3://twine-ci/$TRAVIS_BUILD_NUMBER/ ./coverage --recursive
+        - ls -al ./coverage
+        - bash $TRAVIS_BUILD_DIR/.travis/cc_upload.sh
+      # after_success:
```
astroash: Does this need removing?

.travis.yml
```diff
@@ -17,25 +26,46 @@ services:
...
+  - cd $TRAVIS_BUILD_DIR/$APP_DIR && npm run cover --if-present
 
 after_success:
-  - bash $TRAVIS_BUILD_DIR/.travis/cc-report.sh
+  - bash $TRAVIS_BUILD_DIR/.travis/cc_report.sh
+  - aws s3 cp $TRAVIS_BUILD_DIR s3://twine-ci/$TRAVIS_BUILD_NUMBER/ --recursive --exclude "*" --include "coverage-*.json"
+
+jobs:
+  include:
+    - stage: upload
```
astroash: This job reads as if it only runs on twine-util. Is this correct?

Screen Shot 2019-10-03 at 14 31 02

.travis.yml
eliascodes: oops, should be uncommented

.travis.yml
eliascodes: Apparently this is how you combine matrix expansion with specific individual jobs. IDK why it picks up that `APP_DIR` value, but the upload job doesn't depend on the `APP_DIR` env var, so it doesn't really matter.

PR #334, opened 2019-09-26
♻️ force failure in pre-reqs to avoid having to manually fail in…

eliascodes:
fixes TwinePlatform/twine-api#132

### Changes
- Add `requireTHING` pre-requisites that fail if `isTHING` pre-req returns false

### Testing Requirements
N/A just a refactor

### Release Requirements
N/A

### Manual Deployment
- API

### DIFF
--- a/api/src/api/v1/prerequisites/require_child_organisation.ts
+++ b/api/src/api/v1/prerequisites/require_child_organisation.ts
@@ -5,8 +5,6 @@
  * which is actually a "child" entity.
  *
  * Conditions under which this is true for organisation X:
- * - User is an admin for a funding body which owns at least
- *   one funding programme of which X is a member
  * - User is a Twine admin
  *
  */
--- a/api/src/api/v1/prerequisites/require_child_organisation.ts
+++ b/api/src/api/v1/prerequisites/require_child_organisation.ts
@@ -1,0 +1,26 @@
+/*
+ * "organisations_details-child" route pre-requisite
+ *
+ * Determines if the authenticated user is trying to access an organisation
+ * which is actually a "child" entity.
+ *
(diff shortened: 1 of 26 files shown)


api/src/api/v1/prerequisites/require_child_organisation.ts
```diff
@@ -0,0 +1,26 @@
+/*
+ * "organisations_details-child" route pre-requisite
+ *
+ * Determines if the authenticated user is trying to access an organisation
+ * which is actually a "child" entity.
+ *
+ * Conditions under which this is true for organisation X:
+ * - User is an admin for a funding body which owns at least
```
astroash: "funding body" concept no longer exists. I think we shouldn't include this in our descriptions anymore

api/src/api/v1/prerequisites/require_child_user.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * "user_details-child" route pre-requisite
+ *
+ * Determines if the authenticated user is trying to access an user
+ * which is actually a "child" entity.
+ *
+ * Conditions under which this is true for user X attemptying to access:
+ * user Y
+ * - User X is an admin for a community business for which user Y is a VISITOR
```
astroash: Does this return true for siblings too?
ie 

- User X is an volunteer admin for a community business for which user Y is a volunteer admin

eliascodes: no, only CB admins, check the is_child_user pre-req

astroash: Can this be specified in the description

Issue #338, opened 2019-10-01
📱 Native app implementation


astroash:
- [ ] setup authentication
- [ ] create volunteer app side
- [ ] create volunteer admin side

eliascodes:
Feedback:
- Maybe the settings page can have different subsections just to space out the various fields. So a section for personal info, one for the location stuff, and T&C at the bottom. IMO deffo keep it to one page though.

eliascodes:
Things to remember from feedback on #361:
- [ ] [Check with team](https://github.com/TwinePlatform/twine-monolith/pull/361#discussion_r336532558) about input labels on add time page
- [ ] Date picker [restrictions](https://github.com/TwinePlatform/twine-monolith/pull/361#discussion_r336533338)
- [ ] [Multi-select](https://github.com/TwinePlatform/twine-monolith/pull/361#discussion_r336535921) on admin add time page
- [ ] [Search box](https://github.com/TwinePlatform/twine-monolith/pull/361#discussion_r336535921) on admin add time page
- [ ] Investigate [unFocus](https://github.com/TwinePlatform/twine-monolith/pull/361#issuecomment-543773648) event for projects edit functionality

eliascodes:
From todays meeting:
- Add breakpoints (per day?) in list of logs

astroash:
Note, datepicker just released new version with breaking changes under a `react-native-community`. May be worth upgrading as it comes with min/max date for ios&android

https://github.com/react-native-community/react-native-datetimepicker

astroash:
When close to completion scope fuzzy search for register page cb search

astroash:
- [ ] drop start time & end time
- [ ] readd bulk volunteers

PR #342, opened 2019-10-01
📱 Native app - Volunteer side styles

astroash:
relates #338 

### Changes
- add script for creating component files
- add `native-base` as componenent library
	- `react-native-elements` did not have a picker/dropdown option and was over complicated to make vanilla
- replace all `react-native-elements` componenets
- add Time page
- add Settings page
- setup test suite for CI but no tests yet. Will add when this is being linked up to api
- setup eslint to extend airbnb preset

### Testing Requirements
- [ ] Native app
  - check app builds locally
  - view time page
  - view settings page


nativeApp/bin/touch_tsx/index.ts
```diff
@@ -0,0 +1,44 @@
...
+ * 
+ * FILE: capitalised component name
+ *
+ */
+import getTemplate from './template';
+import { writeFileSync, existsSync, mkdirSync } from 'fs';
+
+const capitaliseFirst = (string) => string.charAt(0).toUpperCase() + string.substring(1);
+
+(() => {
```
eliascodes: Don't think the IIFE is necessary here, the only reason I use it in scripts is because there's no top-level `await` in JS yet.

nativeApp/bin/touch_tsx/index.ts
```diff
@@ -0,0 +1,44 @@
+/*
+ * Command line utility for a creating a blank template for components
+ *
+ * Usage:
+ *  npm run touch:tsx [PATH] [FILE (optional)]
```
eliascodes: Will this really work? Doesn't it need to be `npm run touch:tsx -- PATH [FILE]`

Also `[]` is usually used to indicate optional arguments.

nativeApp/bin/touch_tsx/template.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Types
+ */
+type Props = {
+}
+
+/*
+ * Styles
+ */
+const View = styled.View\`
+  \`;
```
eliascodes: Don't think this is meant to be indented

nativeApp/bin/touch_tsx/index.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ */
+import getTemplate from './template';
+import { writeFileSync, existsSync, mkdirSync } from 'fs';
+
+const capitaliseFirst = (string) => string.charAt(0).toUpperCase() + string.substring(1);
+
+(() => {
+
+  process.on('unhandledRejection', (err) => { throw err; });
```
eliascodes: There's no promises/async stuff in this file so this is unnecessary

nativeApp/bin/touch_tsx/template.ts
```diff
@@ -0,0 +1,31 @@
+export default (fileName) => {
+  return `
+import React, { FC } from 'react';
```
eliascodes: Template includes a blank line at the top of the page

nativeApp/src/lib/ui/forms/Dropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <Label>{label}</Label>
+      <Picker
+        mode="dropdown"
+        iosIcon={<MaterialIcons name="keyboard-arrow-down" />}
+        style={{ width: undefined }}
+        placeholder="Select"
+        placeholderStyle={{ color: ColoursEnum.grey }}
+        placeholderIconColor="#007aff"
+        selectedValue={selectedValue}
+        onValueChange={()=>{}}
```
eliascodes: you don't want to pass the prop here?

nativeApp/src/lib/ui/forms/Dropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+/*
+ * Types
+ */
+type Props = {
+  label: string;
+  options: {id: number; name: string}[];
+  selectedValue?: string;
+  onValueChange?: () => void;
+} & TextInputProps
```
eliascodes: I don't think we should use this form to extend existing types, because it can appear fine on definition, but cause problems on use (in the case where we're overwriting props that already exist)

Example:
```ts
interface One {
  prop: string;
}

type Two = { // this definition looks OK
  prop: number;
} & One;

type Three = One & { // this definition looks OK
  prop: boolean;
}

const x: Two = {
  prop: 1, // Type 'number' is not assignable to type 'never'
};

const y: Three = {
  prop: true, // Type 'number' is not assignable to type 'never'
}
```

Instead if we use `interface` we can see the issue immediately:
```ts
interface Four extends One { // Interface 'Four' incorrectly extends interface 'One'.
  prop: boolean;
}
```

nativeApp/bin/touch_tsx/index.ts
astroash: Yeah it works

nativeApp/src/lib/hooks/requestHook.ts
```diff
@@ -4,18 +4,20 @@ import { AxiosResponse } from 'axios';
...
-      } catch (error) {
-        setError(error)
-      }
-    }()
-  )},[]);
+  useEffect(() => {
+    (
+      async function () {
+        try {
+          const { data: res } = await axiosRequest(payload);
```
eliascodes: isn't it just simpler to use the `Promise` API here?

```ts
useEffect(() => axiosRequest(payload).then((res) => setData(res.result)).catch(setError), [])
```

nativeApp/bin/touch_tsx/index.ts
astroash: Also, I usually use [] to indicate variables rather than optional values. Here the first arg is required n the second is optional. Let me know if you have better way to show this.

nativeApp/src/lib/hooks/requestHook.ts
```diff
@@ -1,21 +1,15 @@
 import { useState, useEffect } from 'react';
-import { AxiosResponse } from 'axios';
 
-const useRequest = (axiosRequest: (x) => Promise<AxiosResponse>, payload?) => {
+const useRequest = (axiosRequest: (x) => Promise<{result: any}>, payload?) => {
```
eliascodes: Lets use a generic here instead of `any`:

```ts
const useRequest = <T, U>(r: (q?: U) => Promise<AxiosResponse<T>>, payload?: U) => {}
```

astroash:
Also published this branch

nativeApp/bin/touch_tsx/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import getTemplate from './template';
+
+const capitaliseFirst = (string) => string.charAt(0).toUpperCase() + string.substring(1);
+
+const [path, file] = process.argv.slice(2);
+if (file) {
+  const fileName = capitaliseFirst(file);
+  writeFileSync(`${path}/${fileName}.tsx`, getTemplate(fileName), { flag: 'wx' });
+} else {
+  const fileName = capitaliseFirst(path.split('/').slice(-1)[0]);
```
eliascodes: We can use `path.basename` here instead, and looking at this made me realise we also need to install `@types/node` otherwise importing the `fs` and `path` modules will cause build errors.

astroash: Is this necessary? From what I can see `path.basename` is an implementation detail, and doesn't offer any additional benefits. Personally, I find template literals easier to read, than unfamiliar node methods. 

With types, this file is a script file so is not imported as part of the build.

Issue #328, opened 2019-09-16
♻️ Models refactor proposal


eliascodes:
# Models refactor proposal
Not especially urgent. This work can be split up into different parts.

## Directory structure
```
|- models
|  |- types
|  |  |- constants.ts
|  |  |- models.ts
|  |  |- queries.ts
|  |  |- records.ts
|  |- collections
|  |  |- users.ts
|  |  |- visitors.ts
|  |  |- volunteers.ts
|  |  |- cb_admins.ts
|  |  |- organisations.ts
|  |  |- community_businesses.ts
|  |  |- temp_community_businesses.ts
|  |  |- volunteer_activities.ts
|  |  |- volunteer_projects.ts
|  |  |- volunteer_logs.ts
|  |  |- visit_activities.ts
|  |  |- visit_logs.ts
|  |  |- single_use_tokens.ts
|  |  |- password_reset_tokens.ts
|  |  |- add_role_tokens.ts
|  |- utils
```

## Collection
Main suggested changes:
- Use function signature overloading to allow methods to make use of more specific types.
- Rename methods to more accurately describe what they do:

A sketch:
```
Collection {
  _toColumnNames;

  // Replaces "create"
  cast;

  // As before but see #for-consideration at the bottom
  serialise;

  // As before
  exists

  // Uses presence of "fields" to determine whether returned object is `Partial` or not
  get (k: Knex, q?: ModelQuery<TModel>): Promise<TModel[]>;
  get (k: Knex, q?: ModelQueryPartial<TModel>): Promise<Partial<TModel>[]>;

  // As with `get`
  getOne (k: Knex, q?: ModelQuery<TModel>): Promise<Maybe<TModel>>;
  getOne (k: Knex, q?: ModelQueryPartial<TModel>): Promise<Maybe<Partial<TModel>>>;

  // Replaces "add": totally generic, DB layer logic.
  // Business logic to be moved to "add" method (e.g. creating roles, etc.)
  create (k: Knex, c: Partial<TModel>): Promise<TModel>;

  // As before
  update

  // Soft-delete
  delete

  // Hard-delete
  destroy
}
```

## Splitting out collections & separating methods
- Volunteer projects collection
- Volunteer activities collection
- Tokens collection(s)
- Visit activities collection
- Visit logs collection
- Temp CB collection
- Separate methods for different parts of visit log aggregation
[...]

Issue #325, opened 2019-09-16
Create v1.1 API


eliascodes:
Related to #282, specifically https://github.com/TwinePlatform/twine-monolith/issues/282#issuecomment-524268263

eliascodes:
Needs consideration for how best to implement, including possibly routing requests at the infrastructure level using reverse proxies.

eliascodes:
@astroash after our discussion on Friday: 

I think this is what you had imagined:
```
/
|- api_v1
|- api_v1.1
|- docs
|- ...
|- volunteer-app
|- ...
```

On reflection, I'm not in favour of this mainly because both these API's would still be sharing the same database. If we completely reproduce the entire repo for each version, there's nothing stopping the database layers of each from changing in inconsistent ways. What if we add a new migration for v1.1 that is backwards incompatible with v1? We'd need to add the migration to the other app too? Then when we re-deploy both apps would want to run the migration? Seems messy.

The API is simply the public interface. It is and should be adequately decoupled from the data layer so that it can change independently from it.

Going with this instead:

```
/
|- api
|  |- src
|  |  |- api
|  |  |  |- v1
|  |  |  |- v1.1
|  |  |- ...
```

Means that both API versions are forced to use the same data layer, ensuring compatibility. Your concern was that changes to the data layer or some utils might then require changes to code in both API versions. This is true, but IMO a price I'm happy to pay if we in return can rely on TS and tests to ensure all versions remain compatible with the database. This strikes me as simply what it means to maintain multiple versions of an API.

PR #345, opened 2019-10-02
♻️ extract serialisers from models

eliascodes:
related to #328 

First step in the refactor outlined in #328, and also the extraction of serialisers is related to projects/logs stuff in #325

### Changes
- Extract serialisers from models

### Testing Requirements
N/A

### Release Requirements
N/A

### Manual Deployment
N/A

### DIFF
--- a/api/src/api/v1/serialisers/index.ts
+++ b/api/src/api/v1/serialisers/index.ts
@@ -1,0 +1,13 @@
+import user from './user';
+import visitor from './visitor';
+import volunteer from './volunteer';
+import volunteerLog from './volunteer_log';
+import communityBusiness from './community_business';
+
+export const Serialisers = {
+  user,
+  visitor,
+  volunteer,
+  volunteerLog,
+  communityBusiness,
+};
--- a/api/src/api/v1/serialisers/index.ts
+++ b/api/src/api/v1/serialisers/index.ts
@@ -1,13 +1,13 @@
-import user from './user';
-import visitor from './visitor';
-import volunteer from './volunteer';
-import volunteerLog from './volunteer_log';
-import communityBusiness from './community_business';
+import * as users from './user';
+import * as visitors from './visitor';
+import * as volunteers from './volunteer';
+import * as volunteerLogs from './volunteer_log';
+import * as communityBusinesses from './community_business';
 
 export const Serialisers = {
-  user,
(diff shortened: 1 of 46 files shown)


api/src/api/v1/serialisers/index.ts
```diff
@@ -0,0 +1,13 @@
+import user from './user';
+import visitor from './visitor';
+import volunteer from './volunteer';
+import volunteerLog from './volunteer_log';
+import communityBusiness from './community_business';
+
+export const Serialisers = {
```
astroash: These method names are very general, each of the items could be serialised in multiple ways. I think it's worth being a bit more specific so we can serve future use cases without having to rename them.

api/src/api/v1/serialisers/volunteer_log.ts
```diff
@@ -0,0 +1,3 @@
+import { identity } from 'ramda';
+
+export default identity;
```
astroash: Is this a functional joke?

eliascodes: No, it's meant as a layer of decoupling. At the moment, our serialisation of logs is just `a -> a`. But if that ever changes it's gonna be easier to replace this function definition than it would be to replace two dozen calls to `Serialisers.identity`. Allows for more specific naming: `Serialisers.volunteerLog` tells you about context, rather than implementation (notwithstanding your comment above, obvs).

api/src/api/v1/serialisers/index.ts
eliascodes: Agree in general, and that was my intent, but at the moment they're all being serialised in the same way.

astroash: How about specifying in the name whether it is removing secrets or nothing?

```
Serialisers = {

  userSecret,
  visitorSecret,
  volunteerSecret,
  volunteerLogNone,
...
```

eliascodes: yeah this is probs better. What about:
```
Serialisers.users.noSecrets
Serialisers.visitors.noSecrets
Serialisers.volunteers.noSecrets
Serialisers.volunteerLogs.identity
```

astroash: Yeah, that works

PR #355, opened 2019-10-14
📱 navigation/style changes & Add Time

astroash:
relates #338 

### Changes
- add specific "Add Time" navigation
- move "Settings" to top left
- add icons to all cards
- make archived projects grey
- create DateTime picker component
- "Add Time" page for volunteers

### Testing Requirements
- [ ] Native app
  - build locally
  - visually inspect new features


nativeApp/src/lib/ui/HoursAndMinutesText.tsx
```diff
@@ -0,0 +1,60 @@
...
+import styled from 'styled-components/native';
+
+import { ColoursEnum } from './colours';
+import { FontsEnum } from './typography';
+
+/*
+ * Types
+ */
+type Props = {
+  timeValues: [number, number];
```
eliascodes: This component's job seems to be to display a duration of time. The representation of that duration shouldn't be affected by how this component chooses to present it.

I think it's better if we allow the parent to pass a more general representation of a duration (e.g. the `Duration` object from `twine-utils`). 

I think we have a couple options:
1. Add `fromDates :: (Date, Date) -> Duration` to `twine-util/duration`
2. Add `reduceTo :: ((keyof Duration)[], Duration) -> Duration` to `twine-util/duration

We then wouldn't need `getTimeDiff` at all, and instead would have:
```ts
type Props = {
  duration: Duration
}

// ...

const [hours, minutes] = Duration.reduceTo(['hours', 'minutes'], props.duration);
```

Or even allow the parent to pass from/to `Date`s and do the conversion to `Duration` internally, if you prefer.

nativeApp/src/screens/admin_views/AddTime/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+
+/*
+ * Component
+ */
+const AddTime: FC<Props> = () => {
+  const [project, setProject] = useState('');
+  const [activity, setActivity] = useState('');
+  const [volunteer, setVolunteer] = useState('');
+  const [date, setDate] = useState();
```
eliascodes: Think we should use `useState<Date>()` on these three, otherwise all the values typecheck as `any`.

nativeApp/src/lib/ui/HoursAndMinutesText.tsx
astroash: I think this is an unnecessary level of abstraction. We are making platform level utilities for app specific styling/functionality.

eliascodes: The intent is the opposite. IMO there is nothing app specific about `fromDates`; it's quite conceivable in general that one would want to generate a `Duration` from two specific dates.

The point is that currently the specific nature of how a duration is presented within `HoursAndMinutesText` is a concern that leaks out of this component, into not only its parent but also into a utility function. That's what I'm trying to avoid.

astroash: Can you explain your approach to component props separation?

For me, I aim to write components as if they were part of an unrelated `component-library`, with their props best representing the way data should be displayed for that visual layout. This aims to separate the specifics of the app by generalising value names and event listeners. Then in page specific components I aim to gather data and manipulate it into the correct shape for various layout components. What you consider as "prop leakage" is something I do by design. 

Here I feel like the use of `Duration` is shoehorning in a complex util that was specifically useful for the dashboard for an app that is concerned with different values and aggregation (or lack of).

nativeApp/src/screens/admin_views/AddTime/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+          mode="date"
+          maxDate={new Date()} // ios only
+        />
+        <DateTimePicker
+          label="Start Time"
+          value={startTime}
+          onConfirm={setStartTime}
+          mode="time"
+        />
+        <DateTimePicker
```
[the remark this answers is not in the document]
astroash: This is visually difficult to show as the placeholder text is already grey.

nativeApp/src/lib/ui/HoursAndMinutesText.tsx
eliascodes: OK I don't want to hold this up tbh, but would also like to clear this up so I've made #360 and am resolving this for now.

Issue #335, opened 2019-09-30
✨ projects/activities page for dashboard


astroash:
Waiting for updated US

**UPDATE**: https://docs.google.com/document/d/1zn2G7vDbOkzawH3eGtrYXItjwW7DYfTaFAqQUsmlFnA/edit

PR #359, opened 2019-10-15
✨ Add projects page to dashboard

eliascodes:
related to #335 

### Changes
- Add projects page
- Update navigation
- NOTE: This **doesn't** include the toggle between projects -> activities which is coming in a subsequent PR

### Testing Requirements
- [x] Dashboard
  - Check navbar displays new working link to `/projects`
  - Check dashboard displays new working link to `/projects`
  - Check graph renders accurately on `/projects`
  - Check table renders accurately on `/projects`
  - Check legend works
  - Check filters/toggles work
  - Check CSV download works and re-creates table in CSV form

### Release Requirements
- Not yet.

### Manual Deployment
- N/A

### DIFF
--- a/dashboard-app/src/features/dashboard/ByProject/datePickerConstraints.ts
+++ b/dashboard-app/src/features/dashboard/ByProject/datePickerConstraints.ts
@@ -1,0 +1,53 @@
+import moment from 'moment';
+import { DateRangePickerConstraint, MIN_DATE } from '../components/DatePicker/types';
+
+// Date rules:
+//
+// - min from: 01 Jan 2017
+// - max from: now
+//
+// - min to: 01 Jan 2017
+// - max to: now
+//
+// - default from: 11 months ago
+// - default to: now
+//
+// - validate:
+//   - from:
+//     - (from < to) || from
+//     - (to - from <= 11 months) || from
+//   - to:
+//     - (from < to) || from
+//     - (to - from <= 11 months) || from + 11 months
+
+
+const TimeConstraints: DateRangePickerConstraint = {
+  from: {
+    min: () => MIN_DATE.toDate(),
+    max: () => moment().startOf('month').toDate(),
+    default: () => moment().subtract(11, 'months').startOf('month').toDate(),
+    validate: (_from, _to) => moment(_from).startOf('month').toDate(),
+  },
+
+  to: {
+    min: (from, to) => moment.max(MIN_DATE, moment(from)).endOf('month').toDate(),
+    max: (from, to) => moment.min(moment(), moment(from).add(11, 'months').endOf('month')).toDate(),
+    default: () => moment().endOf('month').toDate(),
+    validate: (_from, _to) => {
+      const from = moment(_from).startOf('month');
+      const to = moment(_to).endOf('month');
+
+      if (from.isAfter(to)) {
+        return from.endOf('month').toDate();
+      }
+
+      if (to.diff(from, 'months') > 11) {
+        return from.add(11, 'months').endOf('month').toDate();
+      }
+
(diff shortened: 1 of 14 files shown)


dashboard-app/src/features/dashboard/ByProject/datePickerConstraints.ts
```diff
@@ -0,0 +1,53 @@
+import moment from 'moment';
+import { DateRangePickerConstraint, MIN_DATE } from '../components/DatePicker/types';
+
+// Date rules:
+//
+// - min from: 01 Jan 2017
+// - max from: now
+//
+// - min to: 01 Jan 2017
```
astroash: Repeated comment

dashboard-app/src/features/dashboard/ByProject/useAggregateDataByProject.ts
```diff
@@ -0,0 +1,73 @@
...
+      {
+        ...CommunityBusinesses.configs.getVolunteerActivities,
+        transformResponse: [(res: any) => res.result.map(({ id, name }: IdAndName) => ({ id, name }))],
+      },
+    ],
+    updateOn: [...updateOn, from, to],
+  });
+
+  useEffect(() => {
+    if (loading || error) {
```
astroash: Is this a new pattern? Why's this happening?

dashboard-app/src/features/dashboard/ByProject/useAggregateDataByProject.ts
```diff
@@ -0,0 +1,73 @@
...
+      tableType: tableType.ActivityByProject,
+      xData: projectsData.data,
+      yData: activitiesData.data,
+    });
+
+    setAggregatedData(data);
+  }, [logsData, activitiesData, error, loading, projectsData]);
+
+
+  if (loading) {
```
astroash: This will be reached because of the previous return statement

dashboard-app/src/lib/api/index.ts
```diff
@@ -18,12 +18,6 @@ export const Roles = {
 };
 
 export const CbAdmins = {
-  get: () => {
```
astroash: Why has this been removed?

astroash:
Manual testing:
- are we adding stats to the projects box on the home page?

astroash:
Also, it worries me that there are no tests for this data aggregation.

dashboard-app/src/features/dashboard/ByProject/useAggregateDataByProject.ts
eliascodes: This isn't new, it's in all the other `useAggregateData...` hooks too. We don't want that effect to do anything if the data is either loading or there's been an error.

dashboard-app/src/features/dashboard/ByProject/useAggregateDataByProject.ts
eliascodes: It will be (and is) reached. If the previous return statement you're talking about is on line 45, then that return is inside the callback passed to `useEffect`.

dashboard-app/src/lib/api/index.ts
eliascodes: because it's never used.

eliascodes:
> Also, it worries me that there are no tests for this data aggregation.

I've added a test on `ByProjects/index.tsx` in the subsequent PR (#362), which covers the logic in the `useAggregatedDataBy...` hook.

eliascodes:
> are we adding stats to the projects box on the home page?

Yes we should be, although I think I'll add them in a separate PR.

PR #362, opened 2019-10-17
✨ Add toggle between projects/activities to dashboard

eliascodes:
Changes made as part of adding projects chart (see #335)

### Changes
- Add `ProjectsActivitiesToggle` component
- Add ability to prepend additional toggle to `UtilityBar` component
- Switch between two `TableTypeItem` values depending on value of toggle
- Return `yData` directly from `useAggregateDataByProject`

### Testing Requirements
- [x] Dashboard
  - As many combinations of switching back and forth between projects/activities, tables/charts as you can manage

### Release Requirements
- Merge to feature-branch, PO to review feature branch, then release.

### Manual Deployment
- N/A

### DIFF
--- a/dashboard-app/src/App.tsx
+++ b/dashboard-app/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, useState, SetStateAction, Dispatch } from 'react';
+import React, { useState } from 'react';
 import styled from 'styled-components';
 import { Route, Switch, BrowserRouter, withRouter } from 'react-router-dom';
 
@@ -18,6 +18,8 @@ import Navbar from './features/navigation/Navbar';
 import Footer from './lib/ui/components/Footer';
 import { ColoursEnum, Fonts } from './lib/ui/design_system';
 import { DurationUnitEnum } from './types';
+import { DashboardContext } from './features/dashboard/context';
+
 
 /*
  * Styles
@@ -36,18 +38,8 @@ const Content = styled.div`
 
 
 /*
- * Context
- */
-type Context = {
-  unit: DurationUnitEnum;
-  setUnit: Dispatch<SetStateAction<DurationUnitEnum>>;
-};
-export const DashboardContext = createContext<Context>({ unit: DurationUnitEnum.HOURS, setUnit: () => {} });
-
-/*
  * Helpers
  */
-
 
 const DashboardRoutes = () => {
   const [unit, setUnit] = useState(DurationUnitEnum.HOURS);
--- a/dashboard-app/src/features/dashboard/ByActivity/__tests__/index.test.ts
+++ b/dashboard-app/src/features/dashboard/ByActivity/__tests__/index.test.ts
@@ -3,7 +3,7 @@ import MockAdapter from 'axios-mock-adapter';
 import { renderWithHistory } from '../../../../lib/util/tests';
 import { axios } from '../../../../lib/api';
 import { logs, users, activities } from '../../__data__/api_data';
-import ByActivity from '../../ByActivity';
+import ByActivity from '..';
 import 'jest-dom/extend-expect';
 
 
--- a/dashboard-app/src/features/dashboard/ByActivity/index.tsx
+++ b/dashboard-app/src/features/dashboard/ByActivity/index.tsx
@@ -19,7 +19,7 @@ import { getTitleForDayPicker } from '../util';
 import { useErrors } from '../../../lib/hooks/useErrors';
 import { TitlesCopy } from '../copy/titles';
 import { useOrderable } from '../hooks/useOrderable';
-import { DashboardContext } from '../../../App';
+import { DashboardContext } from '../context';
 
 
 /**
--- a/dashboard-app/src/features/dashboard/ByProject/ProjectTabs.tsx
+++ b/dashboard-app/src/features/dashboard/ByProject/ProjectTabs.tsx
@@ -9,7 +9,7 @@ import StackedBarChart from '../components/StackedBarChart/index';
 import { LegendData } from '../components/StackedBarChart/types';
 import { TitleString } from '../components/Title';
 import { Orderable } from '../hooks/useOrderable';
-import { DashboardContext } from '../../../App';
+import { DashboardContext } from '../context';
 
 
 /*
--- a/dashboard-app/src/features/dashboard/ByProject/ProjectTabs.tsx
+++ b/dashboard-app/src/features/dashboard/ByProject/ProjectTabs.tsx
@@ -23,6 +23,7 @@ interface Props {
   title: TitleString;
   legendData: LegendData;
   setLegendData: React.Dispatch<React.SetStateAction<LegendData>>;
+  activeData: string;
 }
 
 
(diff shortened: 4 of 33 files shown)


astroash:
Manual Testing:
- Similar [comment to that on the visitor app](https://github.com/TwinePlatform/twine-monolith/pull/382#issuecomment-552368351). The UX for this toggle is jarring and unclear. This toggle is not a filter, it is changing the datasets & data aggregation. It would make more sense for this to be located elsewhere such as inside the graphs or on a separate line with a description of what it is doing

eliascodes:
> The UX for this toggle is jarring and unclear

I don't agree with this tbh. 

The utility bar is home to several elements, only one of which (the date picker) is a filter, so the toggle not being a filter doesn't really matter.

It's also not changing the data set, it's merely flipping the axes (from `Volunteer Hours` x `Activity` x `Project` to `Volunteer Hours` x `Project` x `Activity`).

I have noticed some other problems though, in the mean-time:
- The graph title doesn't change when toggling between projects/activities
- When toggled to "projects" the table displays only two projects in rows, but when flipped to "activities" the table displays six projects in columns. These should be consistent

This is in addition to the other things you mention.

astroash:
> It's also not changing the data set, it's merely flipping the axes (from Volunteer Hours x Activity x Project to Volunteer Hours x Project x Activity).

Legend aggregation is not an axis. It is changing the structure of the data. If it were flipping axes this would be between `Volunteer Hours` x `Activity`.

That aside, as someone who is unfamiliar with this chart it is unclear what is happening when you toggle this button. There is no description, and when I press it the view completely changes with no visual explanation.

eliascodes:
So do you mean something like this?

Screenshot 2019-11-13 at 15 49 45

astroash:
UX wise that is better, but it could do with a little work on the design and wording.

A couple of suggestions:
- Suggestion One
	- remove the white card background, maybe replace with underline for separation?
	- change wording to something along the lines of "Select to group time by activity or projects"
- Suggestion Two
	- leave toggle in util bar with a "Group by:" label and an ℹ to hover for a full description
- Suggestion Three
	- Place tabs inside other tabs for Project | Activity

Let me know if you would prefer for us to switch tasks for me to put some examples together.

I think it's also worth bringing the rest of the team into the discussion to get others feedback.

eliascodes:
I don't think tabs-within-tabs is a great idea. I'll have a go at the other two suggestions.

eliascodes:
Ok, so:

#### 1. No card, no underline
Screenshot 2019-11-13 at 16 29 55

I think removing the card means we would need to align the toggle with _something_ (though I don't know what), otherwise it looks a little haphazardly placed.

#### 2. No card, underline
Screenshot 2019-11-13 at 16 28 22

I don't think two underlines so close together looks good. As it stands the underline currently seems to serve the visual purpose of separating the controls from the visualisation. Adding another underline within the controls weakens that meaning.

#### 3. Inline, "Group By:" label
Screenshot 2019-11-13 at 16 41 18

I think it looks like the "group by" could apply to both the toggles, even though it only makes sense for the projects toggle.

astroash:
I agree that options 1 & 2 don't look great. I think **option 3** is an improvement, but still not 100% clear for reasons that you have mentioned.

_Going back to tabs..._ I don't think nested tabs are inherently a bad idea.
They'll look something like this:
### Horizontal Tiered Tabs
2 Tier Nav - Admin Main Windows
### Vertical & Horizontal Tabs

tabs




mXMgg

___

Alternatively, we could have 4 tabs on the level named:

`Chart grouped by Project` | `Table grouped by Project` | `Chart grouped by Activity` | `Table grouped by Activity`

eliascodes:
IMO nested tabs look very 90s. None of those examples look good to me. I can't find many favourable discussions of nested tabs online (beyond that they can be acceptable in the right context, where that context is usually some quite complex enterprise software), and interestingly they're also against [the Windows UI Guidelines](https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-tabs).

I've done two more examples of option 3 trying to differentiate the two toggles a bit more:

#### opt 3.1
Screenshot 2019-11-14 at 12 05 42

#### opt 3.2
Screenshot 2019-11-14 at 12 06 43

I think I have a vague preference for 3.2.

I also think that since the position of the toggle was specified in the user story, if we want to have some significant departure from that control, we should take it back to the team rather than spend too long on it here.

astroash:
What's the difference between 3.1 & 3.2? 

Ya, I think it is worth raising these concerns with the rest of the team.

eliascodes:
> What's the difference between 3.1 & 3.2?

The position of the ⓘ -- one is before the toggle the other is after it.

> Ya, I think it is worth raising these concerns with the rest of the team.

OK, so we're deferring that comment on this PR for now? So I'll just address the remaining comments, then move on to Jake's comments in a subsequent PR.

eliascodes:
Summarising the remaining feedback:

> looks like a null project value has slipped in (unless someone created a string value named null)

Should be fixed, push coming.

> switching between Projects/Activities does not keep selected state. Is this wanted behaviour?

By "selected state", do you mean legend data? The legend changes from activities to projects, so not sure how we would persist the state.

> When toggled to "projects" the table displays only two projects in rows, but when flipped to "activities" the table displays six projects in columns. These should be consistent

Looking at it again, this behaviour is consistent with the "volunteers" view -- volunteers with zero hours don't appear in the list. I guess we should probably leave this for now.

> The graph title doesn't change when toggling between projects/activities

I've just changed the title to include both the words "projects" and "activities" since both of them are visible in both views, it's just the aggregation that changes.

astroash:
>By "selected state", do you mean legend data? The legend changes from activities to projects, so not sure how we would persist the state.

Yeah, if I only selected General on `Projects` view, then switch to `Activities` view, then switch back to `Projects` view, the legend selection does not persist.

@eliascodes can you move all of the unresolved comments to a separate issue so they are not lost on merging this PR.

eliascodes:
done #404

astroash:
I'm having the following issue when changing dates. I have reinstalled node_modules
project_issues

eliascodes:
@astroash try building `twine-util`

eliascodes:
There's also a hosted version you can test on

astroash:
Thanks, it was a twine-util issue.

- Is there meant to be a 12 month limit on the time span? I am able to search all of time:
Screen Shot 2019-11-15 at 13 41 42
- the hover time on the ℹ️ is ~2 seconds. No one is going to see this. Is it possible to replace with a js on hover event?
- ℹ️hover on Unit explains length of day, not what the toggle does
- Title is static rather than changing with toggling Projects/Activities
Screen Shot 2019-11-15 at 13 43 03

eliascodes:
> Title is static rather than changing with toggling Projects/Activities

That's deliberate, see [here](https://github.com/TwinePlatform/twine-monolith/pull/362#issuecomment-554327663)

> Is there meant to be a 12 month limit on the time span? I am able to search all of time

That's what the user story specifies.

All your other comments should be addressed.

astroash:
> That's deliberate, see here (bottom-most response)

On second thought if it's static I think the previous title was better. This page shouldn't be drawing attention to activity data.

> That's what the user story specifies.

Can you add this to #404

astroash:
This is looking loads better! Thanks for all the changes ✨

PR #370, opened 2019-10-24
📱 Auth and Redux setup

astroash:
relates #338

### Changes
- refactor Router
- setup authentication
- create Time separator Components
- setup redux
- implement redux examples
- replace `eslint-twine-config` with individual config
- add script to write redux boilerplate
- add script for test boilerplate

### Testing Requirements
- [ ] Native app
  - run locally & confirm it builds

### Release Requirements
- na

### Manual Deployment
- na


nativeApp/src/screens/admin_views/Time/TimePage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {/* TODO - separate logs into separator headers */}
+      <CardSeparator title="Today" />
+      {
+        logs.map((log) => (
+          <TimeCard
+            key={log.id}
+            id={log.id}
+            timeValues={[log.duration.hours || 0, log.duration.minutes || 0]}
+            labels={[log.project || 'General', log.activity]}
+            volunteer={volunteers[log.userId] ? volunteers[log.userId].name : 'Deleted User'}
```
eliascodes: What about:
```ts
pathOr('Deleted User', [log.userId, 'name'], volunteers)
```

nativeApp/src/screens/admin_views/Volunteers/EditVolunteer/index.tsx
```diff
@@ -18,10 +21,24 @@ type Props = {
...
-    <UserForm onSubmit={() => props.navigation.navigate('Volunteers')} />
-  </Page>
-);
+
+// TODO add types
+const EditVolunteer: FC<NavigationInjectedProps<any> & Props> = ({ navigation }) => {
+  const onSubmit = async (formState) => {
+    try {
+      const changeset = filter(Boolean, formState);
+      await CommunityBusinesses.editVolunteer(changeset); // TODO not use ramda
```
eliascodes: Why `TODO not use ramda`?

nativeApp/src/screens/admin_views/Volunteers/VolunteersPage/index.tsx
```diff
@@ -19,27 +27,73 @@ type Props = {
...
-* Component
-*/
+ * Component
+ */
+const Volunteers: FC<NavigationFocusInjectedProps & Props> = ({ navigation }) => {
+  // Redux
+  const dispatch = useDispatch();
 
+  useEffect(() => {
+    dispatch(loadVolunteers());
```
eliascodes: So since we're now loading volunteers on another page, this API call is potentially wasteful. What we want is to understand whether the volunteers that are in our store are outdated. We can add a key `stale` to the state, which is marked `false` by default, but marked `true` if we perform any action which might cause the entity/entities to change on the backend. Then we can check this value before performing another API call. This check could even be part of the thunk (because the component shouldn't really have to care about it).

I think we can make this change in a subsequent PR, rather than needing to do it in this one.

nativeApp/src/screens/admin_views/Volunteers/VolunteersPage/index.tsx
```diff
@@ -19,27 +27,73 @@ type Props = {
...
+  // Handlers
+  const onDelete = (id: number) => {
+    toggleDeleteModalVisibility();
+    setActiveCard(id);
+  };
+
+  const onConfirm = () => {
+    /*
+    * TODO:
+    * - dispatch redux action deleteVolunteerAndUpdate(activeCard.id)
```
eliascodes: This action would be an example of an action that can change the state of the volunteers on the backend, and should therefore mean the list is marked as `stale`.

(Although, alternatively, we know what effect this action has, so we can just remove that user from the store, but I was just highlighting an example of my previous comment).

nativeApp/src/screens/admin_views/Volunteers/VolunteersPage/index.tsx
```diff
@@ -19,27 +27,73 @@ type Props = {
...
         isVisible={deleteModalVisibility}
         title="Delete"
         text="Are you sure you want to delete this volunteer?"
         onCancel={toggleDeleteModalVisibility}
-        onConfirm={toggleDeleteModalVisibility}
+        onConfirm={onConfirm}
+      />
+
+      {volunteersRequestStatus.isFetching && (
+      <ColorDotsLoader
```
eliascodes: This might also be a candidate for a portal? Later though.

eliascodes:
Also I feel like there must be a better/standard way to handle multi-environment configs. Have you come across any libraries that help with this? I found [react-native-dotenv](https://www.npmjs.com/package/react-native-dotenv) but haven't looked seriously.

nativeApp/package.json
```diff
@@ -33,19 +45,20 @@
     "react-navigation": "^4.0.6",
     "react-navigation-stack": "^1.9.0",
     "react-navigation-tabs": "^2.5.5",
-    "styled-components": "^4.3.2"
+    "react-redux": "^7.1.3",
+    "redux": "^4.0.4",
+    "redux-actions": "^2.6.5",
+    "redux-logger": "^3.0.6",
```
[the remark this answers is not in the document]
astroash: I left it commented out in App incase it was wanted during development

nativeApp/bin/utils.ts
```diff
@@ -0,0 +1,4 @@
+export const capitaliseFirst = (string: string) => string
```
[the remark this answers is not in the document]
astroash: I've found file imports intermittently breaking with react native, so if possible would prefer to refrain from using twine-util

nativeApp/src/lib/ui/forms/TimeForm/index.tsx
```diff
@@ -93,7 +95,7 @@ const TimeForm: FC<Props> = (props) => {
         mode="time"
         minDate={startTime}
       />
-      <Label>{`${volunteer || 'Member'} volunteered for`}</Label>
+      <Label>{getTimeLabel(forUser, volunteer)}</Label>
```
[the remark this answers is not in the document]
astroash: It was verbally agreed in a meeting that we would trial this and if users come back with saying that it's an issue we will.

This was an on the spot decision though, it may be worth listing all the small changes introduced in this app for the team to refer back to incase any reasoning changes.

nativeApp/src/screens/admin_views/Time/TimePage/index.tsx
astroash: I was going to avoid using ramda in this app as it has been notoriously annoying with TS. I was planning on writing small util functions or could trial something like lodash if you would prefer.

nativeApp/src/screens/admin_views/Volunteers/EditVolunteer/index.tsx
astroash: See previous comment

nativeApp/src/screens/admin_views/Volunteers/VolunteersPage/index.tsx
astroash: How about loading all entity requests on app load, then any update and deletion actions can manually call a reload after they return?

astroash:
> Also I feel like there must be a better/standard way to handle multi-environment configs.

The current config setup is based on this article:
https://alxmrtnz.com/thoughts/2019/03/12/environment-variables-and-workflow-in-expo.html

nativeApp/src/screens/admin_views/Time/TimePage/index.tsx
eliascodes: I think a subset of ramda is annoying (curry, compose, pipe, evolve and related), but a lot of it is fine, and any problems with inference are usually indicative of type problems in application code.

astroash: `curry` and `compose` is the reason I use ramda. There is no use to a function util library if you cannot use these.

nativeApp/src/screens/admin_views/Volunteers/VolunteersPage/index.tsx
eliascodes: That doesn't seem like it would scale well -- slowing down initial load for data that you might not need is not great. We can discuss this when the next PR is started.

nativeApp/bin/utils.ts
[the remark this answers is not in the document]
eliascodes: I just used it locally and it worked ok

nativeApp/src/screens/admin_views/Volunteers/VolunteersPage/index.tsx
astroash: Yeah, that's a fair point

nativeApp/bin/utils.ts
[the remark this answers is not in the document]
astroash: That's the issue with intermittent bugs...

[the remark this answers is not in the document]
astroash: This is also local script code, so I feel like it is less important than production code. If there turns out to be a bigger need later down the line we can investigate

nativeApp/src/lib/ui/forms/TimeForm/index.tsx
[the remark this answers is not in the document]
astroash: covered in #398

nativeApp/src/screens/admin_views/Volunteers/EditVolunteer/index.tsx
astroash: Was just reminded that the typings on this ramda function meant that it assumed changeset was an array rather than an object, and was breaking types here. This is what prompted my comment

PR #372, opened 2019-10-25
Fix/postcode req

### LINKED ISSUE #371: 🐛 password required
**Issue**
- Password was set as required sign up value, however this is causing issues for 1 cb as they have reported:
> the Postcode field not being optional is problematic because some people don't have set addresses and it's exclusionary.

> it also won't allow duplicate postcodes i.e. same street

(tested and it is possible to use duplicate postcodes)

**Fix**
- remove requirement for postcode sign ups in visitor app
- do not add this requirement for the new volunteer app

astroash:
fixes #371 
### Changes
- visitors endpoint does not require postcode
- visitor app sign up page does not require postcode

### Release Requirements
- ASAP

### Manual Deployment
- Visitor App

### DIFF
--- a/api/src/api/v1/users/register/__tests__/visitors.test.integration.ts
+++ b/api/src/api/v1/users/register/__tests__/visitors.test.integration.ts
@@ -262,6 +262,29 @@ describe('API v1 - register new users', () => {
       }));
     });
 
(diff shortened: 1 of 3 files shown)


astroash:
🚨 self merging

PR #411, opened 2019-11-21
👍 address projects/activities comments

### LINKED ISSUE #404: Remaining UX comments from #362
These can be taken back to the team:

1. Should legend state be persisted when switching back and forth between projects and activities?
2. When toggled to "projects" the table displays only two projects in rows, but when flipped to "activities" the table displays six projects in columns. These should be consistent.
   - although, looking at it again, this behaviour is consistent with the "volunteers" view -- volunteers with zero hours don't appear in the list. Maybe we should probably leave this for now.
3. UX concerns about the projects/activities toggle [see here and following conversation](https://github.com/TwinePlatform/twine-monolith/pull/362#issuecomment-552932406)
4. Double check whether datepickers on each page are actually appropriate settings (i.e. month or day)
5. Double check the time span limit on projects datepicker

eliascodes:
fixes #404 

### Changes
- Add a couple more tests to improve coverage
- Persist legend state when switching back and forth between projects and activities
- Use different colours for charts on projects and activities page
- Get rid of `ProjectsTabs` 
  - This is to avoid passing props through an additional layer unnecessarily
  - The ternary switching the spinner in and out based on the value of `loading` can be replaced with a `LoadingBoundary` component
- `aggregatedToStackedGraphData` takes an optional `GraphColoursEnum[]` argument
- update colours

### Testing Requirements
- [x] Dashboard
  - Projects/Activities page
    - Switch back and forth between projects and activities, while making changes to the legend selection
    - Ensure any changes to a legend are still present when switching back to it
    - Note the colours of the bars in "projects" and "activities" views to ensure they are sufficiently different to avoid confusion

### Release Requirements
- PO Approval of the feature branch

### Manual Deployment
- Dashboard


eliascodes:
Ok so this is "done" except for a small refactor which simplifies the `ByProjects` page component at the expense of introducing an equivalent amount of complexity in the form of a custom hook that handles the switching between canonical data sources.

I think it's OK, but given that I wrote it, it's not clear to me whether it does actually simplify the code or merely obfuscates what's happening.

Hook reproduced below, can push if you like:

```ts
type Arguments<T, U> = {
  initState: T,
  initSide: U,
  left: U,
  right: U
};

export default <T, U> ({ initSide, initState, left: leftSide, right: rightSide }: Arguments<T, U>) => {
  const [left, setLeft] = useState<T>(initState);
  const [right, setRight] = useState<T>(initState);
  const [actual, setActual] = useState<T>(initState);
  const [active, setActive] = useState<(typeof leftSide) | (typeof rightSide)>(initSide);

  const setData = useCallback((x: T | ((y: T) => T)) => {
    if (active === leftSide) {
      setLeft(x);
    } else {
      setRight(x);
    }
  }, [active, leftSide]);

  useEffect(() => {
    if (active === leftSide) {
      setActual(left);
    } else {
      setActual(right);
    }
  }, [active, left, leftSide, right]);

  return {
    active,
    actual,
    setActive,
    setActual: setData,
  };
};
```

If not, we can ignore this and review/merge as is.

astroash:
## Manual Testing:
### Chart Colours
Darker colours look less asthetically pleasing than pastels. Suggest that we use the old range of bright colours to extend colour range
Current dark colour range:
Screen Shot 2019-12-30 at 12 27 56
Screen Shot 2019-12-30 at 12 07 50

Suggested bright colour range:
_note, some tweaking will be needed to separate out similar shades_

Screen Shot 2019-12-30 at 12 27 02

Screen Shot 2019-12-30 at 12 19 50

Issue #409, opened 2019-11-19
expand gender options


eliascodes:
list to be provided by @edwardwptc

edwardwptc:
Let's go with:
Male
Female
Non-binary
Intersex
Other
Prefer not to say

Don't read this if you're not interested but here's our rationale: 
Intersex people, those with not clearly male or female primary sex characteristics at birth, are a statistically significant portion of the population at birth (between 1.5 to 5%). This can be due to genetic disorders, chromosomal disorders, or pure physical 'abnormalities'. Statistics are uncertain because Western society doesn't recognise this as "normal" and parents/doctors tend to make decisions about the child's sex one way or the other. As such, the adult population who are intersex is much lower. Many may be living without ever knowing they were intersex at birth. 

The general thinking is that "Intersex" is a category of "Sex". Male, Female, Intersex, in terms of biology.  There's a growing movement which suggests Intersex is an appropriate gender category too - people can identify as intersex. It's different to all the other categories, so we'll list it :)

astroash:
> Don't read this if you're not interested but here's our rationale:

Gender diversity is something that interests me. I've not come across intersex described as a gender before but totally get that people may choose this term to identify with.

On defaults, I assume:
- we will be adding these extra options
- gender will still be optional
- if not selected will default to 'prefer not to say'

edwardwptc:
Assumptions are perfect. 

Further reading about the intersex as a gender thing if you're interested:
https://www.intersexequality.com/how-intersex-people-identify/

astroash:
That was an interesting read. Thank you for sharing 🙂

PR #412, opened 2019-11-22
🧩 expand gender options in validation schema

eliascodes:
related to #409 

### Changes
- Adds items detailed in #409 to list of permitted gender options for API payloads

### Testing Requirements
- [x] API
  - Send `POST /v1/users/register/visitors` with one of the new gender options
  - Expect 200

### Release Requirements
- When ready, no PO approval required

### Manual Deployment
- API
- After deployment, add new gender options to the database with:
```sql
insert into gender (gender_name) values ('non-binary'), ('intersex'), ('other');
```


astroash:
Unclear why genders are being added manually rather than with migrations. Will add as migration for completeness.

PR #382, opened 2019-11-04
✨ update visits charts page

eliascodes:
fixes #324 

### Changes
- Use newly extracted components and code (for charts, CSV and data processing)
- Toggle between visits and visitor data
- Tab between demographics and numbers data
- Add "All" as the "clear filter" value
- User can select date range for data
- Dynamically display the date range for the data

### Testing Requirements

NB review feedback being moved to #413 

Link to Requirements in #324
- [ ] Visitor
  - Toggle
    - Can toggle between _visits_ and _visitors_ data
  - Date range filter
    - Data changes to accurately reflect time range selected
  - Age filter
    - Data changes to accurately reflect age range selected
  - Activity filter
    - Data changes to accurately reflect activity selected
  - Export button
    - Allows download of CSV of data that reflects the filter settings
  - Tabs
    - Can tab between demographics data (doughnut charts) and numbers data (bar charts)
  - Charts
    - Can click on categories charts to drill down into activities, and then click again to return to top level

### Release Requirements
- Merge to branch

### Manual Deployment
n/a

### NOTES
- ~`react-flexbox-grid` will be replaced shortly, just used it for convenience in order to alter layout quickly for demo.~
- Have an activity filter. Activities are user defined. Do we want a category filter instead?
- CSV export currently does not respect time range. Should it?

PR #413, opened 2019-11-27
🚸 improve UI/UX on visit charts page

### LINKED ISSUE #394: Visits charts feedback
We received from feedback from Jake and Stephen.

> In the dropdown list for 'Activities', what does 'project activity' mean?

Everything in that list is a user defined activity. You can confirm this by going to the activities page in the admin section.

> The colours switch in the charts when applying filters. For example, male goes from purple to yellow when filtering by Gender or Activity. Same thing happens to the Age chart too. If possible, would like us to keep consistent colours

To be actioned.

> We also previously discussed using the same colour palette as the vol app chart, which can handle multiple categories. Has this been applied? It's not clear how the demographics charts looks when loaded with multiple age groups

To be addressed in a separate issue. See #395.

> Visits by time chart would be better above the 'reasons for visiting' chart if possible, as this is probably the more useful info for CBs

To be actioned.

> Could we align the Y axes on the two charts? I think it might look slightly neater than having them offset from one another

Possibly. Will investigate if this is functionality that is exposed by our chart library.

> On the ‘Reasons for Visiting’ chart, where an activity has only 1 or 2 entries (as is the case for ‘Skill share’ and ‘Digital share’ – the bar is very small. Are there any small adjustments we can make to the scale to increase those sizes or are we limited by the upper range numbers?

If I follow what you're saying, are you suggesting that we have a non-linear x-axis?

> The charts run very close to the bottom of the screen, and it makes it look like there might be more to see below the fold – could we add a little more padding under the charts at the bottom?

To be actioned.

> Add initial caps for “Male”, “Female” and “Prefer not to say” so we are consistent across our labelling

To be actioned.

> ‘Prefer not to say’ data is in grey, but we also have a small white separation between sections on the pie charts, which makes it a little difficult to see. Could we choose a different third colour? One which more contrast to the white?

To be actioned. Related to colour palette selection in general.

> Not been able to login on Chrome (though this might have been after you changed the password) – what browsers are we using for testing?

We currently only explicitly support Chrome, but it should work fine on Firefox too (my main browser). Safari sometimes has problems on the QR scanning page. We've never tried it on IE/Edge.

I've been able to log in on both Chrome and Firefox, so I imagine it'll just be because I changed the password. Do you have access to LastPass?

> Export function not working for me – is it supposed to work on this version?

It should work. In what way is it not working?

> I’m not sure we need to have “Dates:” before the date range? I think we can just display the dates themselves

To be actioned.

### Summary of actions on dev
- [x] Remove "Dates:"
- [x] Darken "Prefer not to say" colour
- [x] Consistent capitalisation of gender labels
- [x] Add padding to bottom of charts
- [x] Swap "Reasons for visiting" chart with "Visits by time"
- [x] Prevent colours switching when changing applied filters

### LINKED ISSUE #395: 🎨 Colours on the visits charts
Based on feedback in #394

> We also previously discussed using the same colour palette as the vol app chart, which can handle multiple categories. Has this been applied? It's not clear how the demographics charts looks when loaded with multiple age groups

Assuming you mean "the same colour palette as the dashboard charts". 

Are you suggesting using this palette for the bar charts, the doughnut charts, or both? 

I think the bar charts would not be improved by using colours other than the Twine purple.

I've attached shots of a number of different options:

### Looping over the same 3 colours
Screenshot 2019-11-07 at 16 31 07

### Colour palette option 1
Screenshot 2019-11-07 at 16 38 53

### Colour palette option 2
Screenshot 2019-11-07 at 16 39 39

### Colour palette option 3
Screenshot 2019-11-07 at 16 40 27

### Colour palette option 4
Screenshot 2019-11-07 at 16 41 09

### Colour palette option 5
Screenshot 2019-11-07 at 16 41 43

Personally, I'm not sure introducing new colours outside of the yellow/purple/grey/white colour palette of the visitor app wins us much legibility beyond the first option of looping the three colours. However my sense of design is not great, and I may be discounting people who find charts harder to read than I do, so I'm happy to go with whatever your preferences are.

eliascodes:
**DO NOT MERGE**
**MERGE #382 FIRST**

fixes #394 
fixes #395 

### Changes
- Assign fixed colours to chart gender and age categories
- Capitalise genders
- Move reasons for visiting chart to bottom
- Add bottom spacing
- Use colour palette from design system on doughnut charts to accommodate more than 3 categories

### Testing Requirements
- [x] Visitor app -- Visits page
  - Ensure colours for each category remain consistent when changing filter values
  - Ensure genders are consistently capitalised

### Release Requirements
- PO approval, release with feature branch

### Manual Deployment
- Visitor app

### DIFF
--- a/visitor-app/src/cb_admin/pages/VisitsData/__tests__/data.test.js
+++ b/visitor-app/src/cb_admin/pages/VisitsData/__tests__/data.test.js
@@ -92,9 +92,8 @@ describe('Visits Dashboard Data Processing', () => {
             labels: ['Male', 'Female'],
           },
           time: {
-            datasets: [{ data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2], backgroundColor: '#833FF7', label: undefined }],
+            datasets: [{ data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2], backgroundColor: '#833FF7', label: undefined }],
             labels: [
-              'Oct 2018',
               'Nov 2018',
               'Dec 2018',
               'Jan 2019',
--- a/visitor-app/src/cb_admin/pages/VisitsData/__tests__/data.test.js
+++ b/visitor-app/src/cb_admin/pages/VisitsData/__tests__/data.test.js
@@ -74,7 +74,7 @@ describe('Visits Dashboard Data Processing', () => {
         },
         data: {
           age: {
-            datasets: [{ data: [1, 1], backgroundColor: ['#FDBD2D', '#833FF7', '#DBDBDB'], label: undefined }],
+            datasets: [{ data: [1, 1], backgroundColor: ['#6370BC', '#379A90'], label: undefined }],
             labels: ['18-34', '51-69'],
           },
           activity: {
@@ -88,8 +88,8 @@ describe('Visits Dashboard Data Processing', () => {
             },
           },
           gender: {
-            datasets: [{ data: [1, 1], backgroundColor: ['#FDBD2D', '#833FF7', '#DBDBDB'], label: undefined }],
-            labels: ['male', 'female'],
+            datasets: [{ data: [1, 1], backgroundColor: ['#47ABFA', '#6370BC'], label: undefined }],
+            labels: ['Male', 'Female'],
           },
           time: {
             datasets: [{ data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2], backgroundColor: '#833FF7', label: undefined }],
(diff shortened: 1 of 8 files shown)


astroash:
## Manual Review
Carried over from #382 
### Toggle
- [x] Chart titles do not change to reflect visit/visitors toggle

### Date Range Filter 
- [x] Date range appears to be 1 year and 1 day eg "30 Dec 2018 - 30 Dec 2019".

### Export button
- [ ] Date range is not reflected in CSV export, but does for filters. Suggest appending date range eg visits_data_18-12-30_to_19-12-29
Check with PO to confirm

astroash:
Moving final review point to FEATURE branch PR

PR #475, opened 2020-07-18
merge to get the new editTime table

kyeap:
fixes ...

### Changes
- 
-

### Testing Requirements
- [ ] 
  - 
  - 

### Release Requirements
- 

### Manual Deployment
-


kyeap:
Merge so that error trapping is also in volunteerTime page development. Due to the conflict of TimeForm page, I decided to not pass the duration as one of the props from editTime page. Currently TimeForm page is used by EditTime and AddTime, when from EditTime props project and activity is passed. Since the new TimeForm page uses time instead of duration in the dropdown. 

Therefore, currently, the duration will not be passed and in the future, it can be considered to make it a better UX to include the time conversion from:
- duration -> date
- from starting date 
- till starting date + duration 
form the date existing code will convert it back to duration and send with the api.

